PR #61, opened 2017-10-17
Tooling

ycleptkellan:
Introduces the following to define and maintain consistent coding style:
 -  an [EditorConfig](http://editorconfig.org/)
 - [TSLint](https://palantir.github.io/tslint/) ([react](https://github.com/palantir/tslint-react), [eslint-recommended](https://github.com/buzinas/tslint-eslint-rules)) 
 - [Prettier](https://github.com/prettier/prettier) 💄 to keep things tidy

Adds [Hot Module Replacement](https://webpack.js.org/concepts/hot-module-replacement/) (HMR) by way of [React Hot Loader](https://github.com/gaearon/react-hot-loader) so that we can maintain state while editing components. 

Note: TSLint and Prettier are obv opinionated by nature, happy to go back and forth on what we do/don't want to enforce. And if we want to take things to the next level, I vote we consider [tslint-immutable](https://github.com/jonaskello/tslint-immutable).


ycleptkellan:
nice, thanks

PR #62, opened 2017-10-19
Jest in time

ycleptkellan:
- Adds Facebook’s [Jest](https://facebook.github.io/jest/) testing platform
 - Adds support for TypeScript preprocessing/sourcemap support via [ts-jest](https://github.com/kulshekhar/ts-jest) ([though we could also write our own](https://github.com/facebook/jest/tree/master/examples/typescript))
 - Adds [enzyme-to-json](https://github.com/adriantoine/enzyme-to-json), which converts enzyme wrappers to a format compatible with Jest snapshot testing. It also appears to improve snapshot readability. 
- Moves test files next to the components they’re responsible for testing (see `utils.test.ts`. Opinions vary on test organization (e.g. [here](https://medium.com/@JeffLombardJr/organizing-tests-in-jest-17fc431ff850) and [here](https://medium.com/selleo/testing-react-components-best-practices-2f77ac302d12)), but most seem to at least favor placing unit tests next to their components.
 - Bumps enzyme to v3 to accommodate enzyme-to-json
 - Introduces [jest-enzyme](https://github.com/blainekasten/enzyme-matchers), which provides React-specific assertion helpers and more verbose/contextual assertion errors than Jest's default helpers. 
 - Moves `tsx`-absent files to the `ts` extension. Proposing that all `ts` files that don't contain `tsx` move to `.ts` extension. 
 - Getting a feel for each component and adding optimizations/fixing linting errors where I can.


src/components/FlowLoader.tsx
```diff
@@ -51,36 +45,38 @@ export class FlowLoader extends React.PureComponent<FlowLoaderProps, FlowLoaderS
...
-        var flow = null;
-        if (this.state.definition) {
-            flow = <Flow
-                definition={this.state.definition}
-                dependencies={this.state.dependencies}
-                mutator={this.mutator} />
-        }
-        return flow;
+        return (
+            this.state.definition ? (
```
ericnewcomer: Is this a code formatter forcing a ternary here?

src/services/External.ts
```diff
@@ -0,0 +1,201 @@
...
+                .get(this.endpoints.flows, { headers: this.headers })
+                .then((response: AxiosResponse) => {
+                    resolve(response.data.results as FlowDetails[]);
+                })
+                .catch(error => reject(error));
+        });
+    }
+
+    /** Filter WIP */
+    // public getFlows(name: string): Promise<FlowDetails[]> {
```
ericnewcomer: Why did this get commented out?

ericnewcomer:
Ah, I see this is all very much WIP.

- Snapshot testing is interesting, I think I like that
- Tests by components works for me
- Not sure .ts buys us anything, but fine either way there
- Still have some reservations on some of the forced formatting that I don't really agree with (removing some organizational options), but will take some I don't like over no forced formatting, so happy to stay the course there for now

Approach so far seems sane, would like to see this continue toward testing an actual component with a coverage report and TS line numbers. It's possible that we might need to rework things to make them more testable, all the more reason to tackle this now though.

src/components/FlowLoader.tsx
ycleptkellan: yes, but happy to remove

ycleptkellan:
Aye! I'll continue in that direction and lay off the formatting for now.

ericnewcomer:
To be clear, I'm cool with the formatting stuff, just don't want it to force a ternary.

ycleptkellan:
Pssh, who doesn't love ternaries?! 

I'm running into an issue [here](https://github.com/nyaruka/floweditor/blob/jest-in-time/src/components/FlowLoader.tsx#L55) while attempting to test render on `FlowLoader` whereby Enzyme's `shallow` method appears to be triggering `FlowLoader`'s `componentDidMount` lifecycle callback and thus erroring out because `Config` is initialized higher up the render chain in `Editor` and therefore `undefined` in `FlowLoader`. This might be a good place to have a discussion. At first blush, it might make sense to initialize `Config` in `Editor`'s `constructor` and pass it down to `FlowLoader` via props). Not sure it makes sense to push it up to `Editor`'s state, unless it's likely to change while a user's interacting with the editor. What do you think?

PR #67, opened 2017-11-09
Use jest-enzyme matchers instead of global type defs

ericnewcomer:
This fixes the warnings in vscode for enzyme matchers
screen shot 2017-11-09 at 11 53 46 am


ycleptkellan:
awesome thanks!

PR #80, opened 2017-12-02
Localization pre population

### LINKED ISSUE #73: Localization editor shouldn't be pre-populated
We shouldn't pre-propulate the translation when it is empty

image 2017-12-01 13-52-42

ycleptkellan:
Resolves #73

### DIFF
--- a/src/components/Node.tsx
+++ b/src/components/Node.tsx
@@ -2,10 +2,18 @@ import * as React from 'react';
 import { Language } from './LanguageSelector';
 import Action, { ActionProps } from './Action/Action';
 import { IDragEvent } from '../services/Plumber';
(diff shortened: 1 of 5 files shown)


src/components/Node.tsx
```diff
@@ -261,7 +270,7 @@ export default class NodeComp extends React.Component<NodeProps, NodeState> {
                 }
             }
 
-            // add our exit localizations
+            /** Account for localized exits */
```
ericnewcomer: Curious the purpose for this commenting style

ycleptkellan: We use jsDoc comments elsewhere so just trying to make things consistent. Not something I'm passionate about either way.

PR #95, opened 2017-12-22
Dnd style

### LINKED ISSUE #92: Small resize on DnD
When dragging rules, the drag space underneath shrinks a few pixels, then when dropping it gains those pixels bank. To fix this visual tick, we should make sure the spacing of the hover container is the same height so that the parent container height doesn't change when moving things around.

ycleptkellan:
resolves #92


ycleptkellan:
http://gph.is/2zhlhL9

ycleptkellan:
yikes, coverage forthcoming

ycleptkellan:
Ah ok I'll let you know when that's fixed

PR #96, opened 2018-01-09
Fix dnd style

### LINKED ISSUE #92: Small resize on DnD

ycleptkellan:
resolves #92, breaks out helper methods and adds tests. 

http://gph.is/2zhlhL9


src/components/actions/Reply/ReplyForm.tsx
```diff
@@ -31,7 +31,7 @@ export default class ReplyForm extends React.Component<ReplyFormProps> {
     public onValid(widgets: { [name: string]: any }): void {
         const localizedObject = this.props.getLocalizedObject();
 
-        const textarea = widgets['Message'] as TextInputElement;
+        const textarea = widgets.Message as TextInputElement;
```
ericnewcomer: this feels weird, would much rather they not be accessed with dot notation since they are keyed by name which has case and spaces, etc.

ycleptkellan: OK tslint was yelling at me over that: https://palantir.github.io/tslint/rules/no-string-literal/

Case does make it awkward, happy to turn off that rule & revert.

ericnewcomer: I want my cake too though. No argument, I definitely don't like implicit any on normal object properties, but don't want to have that as an issue for dictionary lookups. The explicit cast here I thought should address this though.

ycleptkellan:
I haven't been able to replicate that for some reason, can I have a repro case to try?

src/components/actions/Reply/ReplyForm.tsx
ycleptkellan: Gotcha, I agree but it looks like they're just checking the lookup string: 

https://github.com/palantir/tslint/blob/4c29dd9fd14afda79ab8e2b2fc10a868adbc9e44/src/rules/noStringLiteralRule.ts#L54 

That might be why it yells even with the type cast.

PR #102, opened 2018-01-09
Update localization

ycleptkellan:
- forces creation of `LocalizedObjects` in `Node` component regardless of whether the user is actively translating (fixes a bug I introduced in a previous commit). 
- threads through and leans on `language` prop to determine app's current language state. 
- bit of housekeeping within components affected.


src/helpers/utils.ts
```diff
@@ -15,45 +15,41 @@ interface BoolMap {
  * @param {string[]} array - an array of strings, e.g. contact fields
  * @returns {object} A map of each string
  */
-export function toBoolMap(array: string[]): BoolMap {
-    return array.reduce(
+export const toBoolMap = (array: string[]): BoolMap =>
```
ericnewcomer: Why the switch here? Don't see why we wouldn't want utils to be proper functions. I get that it'll all be pre-imported anyways, but willing to be enlightened here.

ycleptkellan: Function declarations are hoisted but stuffing them into variables (function expressions) removes that badness. One could argue it doesn't matter in that file but why not set a precedent and rule them out 🤷‍♂️

ericnewcomer: Right, the thinking there is that util functions are kinda global things, so hoisting seems appropriate. But whatevs, no strong feelings here either way.

PR #120, opened 2018-01-17
Add/remove groups

### LINKED ISSUE #7: Groups should support multi-select
Including Add to Group, Remove from Group, Split by Group, etc

ycleptkellan:
- Closes #7, #127 
- Adds 'remove from all groups' field/action
- Improves form field invalid entry/error feedback for add/remove group forms, others
- Adds `getAction` method to `NodeEditor.tsx` to ensure each action form has an action when it's rendered. This removes the need to check for the presence of an action in each action/hybrid form.


ycleptkellan:
@ericnewcomer this is G2G

ycleptkellan:
@ericnewcomer I split `ChangeGroupForm` in two to get around `NodeEditor` losing the 'Group' widget ref when switching between the add and remove group forms. Similar to this: 

https://github.com/nyaruka/floweditor/issues/123

src/components/form/FormElement.tsx
```diff
@@ -1,78 +1,82 @@
...
-    } else {
-        helpText = '';
-    }
+        const errors: JSX.Element[] = this.props.errors.map((error, idx) => {
+            const className = cx({
+                [styles.error]: true,
+                [styles.replyError]: this.props.replyError === true,
+                [styles.kaseError]: this.props.kaseError === true,
+                [styles.fieldError]: this.props.fieldError === true
+            });
```
ericnewcomer: What does this produce? Is it a single merged class or somesuch? Just curious what using this library is buying us.

src/components/form/TextInputElement/TextInputElement.tsx
```diff
@@ -84,21 +85,30 @@ type InitialState = Pick<
...
+const KEY_TAB: number = 9;
+const KEY_P: number = 80;
+const KEY_N: number = 78;
+const KEY_ESC: number = 27;
+const KEY_BACKSPACE: number = 8;
+
+export const MAX_GSM_SINGLE: number = 160;
+export const MAX_GSM_MULTI: number = 153;
+export const MAX_UNICODE_SINGLE: number = 70;
+export const MAX_UNICODE_MULTI: number = 67;
```
ericnewcomer: Isn't number implicit here?

src/components/form/FormElement.tsx
ycleptkellan: Right, exactly it just returns a string containing the classes that will be included. Thought it might be worth introducing to cut down on class concatenation logic in `render()`.

src/components/form/TextInputElement/TextInputElement.tsx
ycleptkellan: doh, yeah, I got into the habit of annotating everything at some point. happy to remove

src/components/routers/SubflowRouter.tsx
```diff
@@ -109,24 +108,15 @@ export default class SubflowRouter extends React.PureComponent<SubflowRouterForm
...
             <div>
                 <p>Select a flow to run</p>
                 <FlowElement
                     ref={this.props.onBindWidget}
                     name="Flow"
                     endpoint={this.context.endpoints.flows}
-                    flow_name={flowName}
-                    flow_uuid={flowUUID}
+                    flow_name={(this.props.action as StartFlow).flow_name}
+                    flow_uuid={(this.props.action as StartFlow).flow_uuid}
```
ericnewcomer: Having two casts here feels odd to me. Since we know we have a StartFlow action, maybe we make that a first class thing that's typed?

src/components/routers/WebhookRouter.tsx
```diff
@@ -356,20 +353,17 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
                 }
             }
         }
-
         var newAction: CallWebhook = {
-            uuid: this.props.getActionUUID(),
+            uuid: this.props.action.uuid,
```
ericnewcomer: Without looking, I'm guessing we've now guaranteed that action.uuid is there whereby the function was conditionally determining it?

ycleptkellan: that's exactly right, the `getAction` method on `NodeEditor` is guaranteeing that now

src/components/routers/SubflowRouter.tsx
ycleptkellan: Ah right, that's an oversight on my part. Changing the `action` prop in the `SubflowRouterFormProps` interface to `StartFlow` removes the need to cast there.

src/helpers/utils.ts
```diff
@@ -107,3 +107,26 @@ export const reorderList = (list: any[], startIndex: number, endIndex: number):
...
     JSON.stringify(objA) === JSON.stringify(objB);
+
+/**
+ * Checks whether any of a list of error strings contain one or more queries.
+ * Used to determine whether a particular error has been encountered.
+ * @param {Array.<string>} errors - list of error messages
+ * @param {Array.<string>} queries  - list of queries
+ * @returns {boolean}
+ */
+export const hasErrorType = (errors: string[], queries: string[]): boolean => {
```
ericnewcomer: Not sold on this, but willing to give it a go

PR #125, opened 2018-01-22
PUT option

ycleptkellan:
- refactors `WebhookRouter` to appease tslint
- adds `Methods` enum 
- adds `PUT` option


ericnewcomer:
Can't add it to the editor without also adding an issue for it on goflow

ycleptkellan:
doh will do

ycleptkellan:
Do you think the `Methods` enum is enough validation for now? 

https://github.com/nyaruka/goflow/issues/143

ericnewcomer:
Ya, as long as it's restricted to the set we want, I think we are good here.

PR #137, opened 2018-02-13
fix headers

ycleptkellan:
fixes issue whereby headers weren't displaying properly for nodes that don't already contain webhook routers

### DIFF
--- a/src/components/NodeEditor/NodeEditor.tsx
+++ b/src/components/NodeEditor/NodeEditor.tsx
@@ -436,6 +436,8 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
      * Returns existing action (if any), or a bare-bones representation of the form's action.
      */
     private getAction(): AnyAction {
+        let uuid: string;
+
         if (this.props.action) {
             if (
                 this.props.action.type === this.state.config.type ||
@@ -443,44 +445,56 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
                     this.props.action.type === this.state.config.aliases[0])
             ) {
                 return this.props.action;
+            } else {
+                ({ uuid } = this.props.action);
             }
         }
 
-        let action: Action = {
+        let defaultAction: Action = {
             type: this.state.config.type,
-            uuid: generateUUID()
+            uuid: uuid || generateUUID()
         };
 
         switch (this.state.config.type) {
             case 'reply':
-                action = { ...action, text: '', all_urns: false } as Reply;
+                defaultAction = { ...defaultAction, text: '', all_urns: false } as Reply;
                 break;
             case 'add_to_group':
-                action = { ...action, groups: null } as ChangeGroup;
+                defaultAction = { ...defaultAction, groups: null } as ChangeGroup;
                 break;
             case 'save_contact_field':
-                action = {
-                    ...action,
+                defaultAction = {
+                    ...defaultAction,
                     field_uuid: generateUUID(),
                     field_name: '',
                     value: ''
                 } as SaveToContact;
                 break;
             case 'send_email':
-                action = { ...action, subject: '', body: '', emails: null } as SendEmail;
+                defaultAction = {
+                    ...defaultAction,
+                    subject: '',
+                    body: '',
+                    emails: null
+                } as SendEmail;
                 break;
             case 'save_flow_result':
-                action = { ...action, result_name: '', value: '', category: '' } as SaveFlowResult;
+                defaultAction = {
+                    ...defaultAction,
+                    result_name: '',
+                    value: '',
+                    category: ''
+                } as SaveFlowResult;
                 break;
             case 'call_webhook':
-                action = { ...action, url: '', method: 'GET' } as CallWebhook;
+                defaultAction = { ...defaultAction, url: '', method: 'GET' } as CallWebhook;
                 break;
             case 'start_flow':
-                action = { ...action, flow_name: null, flow_uuid: null } as StartFlow;
+                defaultAction = { ...defaultAction, flow_name: null, flow_uuid: null } as StartFlow;
                 break;
         }
 
-        return action;
+        return defaultAction;
     }
 
     private getMode(): Mode {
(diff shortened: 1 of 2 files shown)


src/components/NodeEditor/NodeEditor.tsx
```diff
@@ -436,51 +436,65 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
...
                     this.props.action.type === this.state.config.aliases[0])
             ) {
                 return this.props.action;
+            } else {
+                ({ uuid } = this.props.action);
             }
         }
 
-        let action: Action = {
+        let defaultAction: Action = {
```
ycleptkellan: Ensuring we're using the existing action's `UUID` if applicable. Important for `FlowMutator`'s ' `updateAction` method

src/components/NodeEditor/NodeEditor.tsx
```diff
@@ -436,51 +436,65 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
...
+                defaultAction = { ...defaultAction, url: '', method: 'GET' } as CallWebhook;
                 break;
             case 'start_flow':
-                action = { ...action, flow_name: null, flow_uuid: null } as StartFlow;
+                defaultAction = { ...defaultAction, flow_name: null, flow_uuid: null } as StartFlow;
                 break;
         }
 
-        return action;
+        return defaultAction;
```
ycleptkellan: i think `defaultAction` makes more sense here

src/components/routers/WebhookRouter.tsx
```diff
@@ -123,7 +123,7 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
             return map;
         }, {});
 
-        const uuid: string = (this.props.action && this.props.action.uuid) || generateUUID();
+        const { uuid } = this.props.action;
```
ycleptkellan: No need to check for an existing action or a `UUID` on that action since `NodeEditor`'s `getAction` method ensures we're provided with a `UUID`.

src/components/routers/WebhookRouter.tsx
```diff
@@ -244,24 +244,24 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
             resultName: null,
             setResultName: false,
             cases: [],
+            headers: [{ name: '', value: '', uuid: generateUUID() }],
```
ycleptkellan: includes an empty header by default

src/components/routers/WebhookRouter.tsx
```diff
@@ -244,24 +244,24 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
...
-                    }))
-                );
+            if (webhookAction.headers && Object.keys(webhookAction.headers).length) {
+                const existingHeaders = Object.keys(webhookAction.headers).map(key => ({
+                    name: key,
+                    value: webhookAction.headers[key],
+                    uuid: generateUUID()
+                }));
+
+                initialState.headers.unshift(...existingHeaders);
```
ycleptkellan: adding to our headers list if existing action contains headers

src/components/routers/WebhookRouter.tsx
```diff
@@ -290,7 +290,7 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
         let method = Methods.GET;
         let url = '';
 
-        if (this.props.action && this.props.action.type === 'call_webhook') {
+        if (this.props.action.type === 'call_webhook') {
```
ycleptkellan: action exists per `NodeEditor`'s `getAction` method

src/components/routers/WebhookRouter.tsx
```diff
@@ -305,18 +305,18 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
         const linkText = this.state.method === Methods.GET ? baseText : `${baseText} and body`;
 
         return (
-            <span>
+            <React.Fragment>
```
ycleptkellan: Constants and Fragments can be considered stylistic changes but I'd argue this is important housekeeping since constants improve code readability/allow us to reuse values (esp in tests) and Fragments reduce the amount of HTML output to the page.

src/components/routers/WebhookRouter.tsx
```diff
@@ -273,7 +273,7 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
         let hasEmpty = false;
 
         for (const header of newHeaders) {
-            if (header.name.trim().length === 0 && header.value.trim().length === 0) {
+            if (!header.name.trim().length && !header.value.trim().length) {
```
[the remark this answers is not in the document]
ericnewcomer: Not saying don't do this. Just sharing my bias is generally to avoid truthy/falsy and defer to more explicit conditions. While it is equal here, experience as a practice is this can lead to unintended consequences when it really isn't buying you anything. ie, you may not want munge the handling of false, 0, NaN, and null, but might not consider them consistently with this practice.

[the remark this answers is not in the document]
ericnewcomer: Also, I believe HTTP headers can have empty values, so perhaps we should allow named headers without a value set.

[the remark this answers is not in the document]
ycleptkellan: gotcha, will revert. I believe we do allow for headers to contain a name but no value per `HeaderElement`'s `validate` method.

src/components/routers/WebhookRouter.tsx
```diff
@@ -335,8 +335,8 @@ export default class WebhookForm extends React.Component<WebhookRouterFormProps,
         const summary: JSX.Element = this.getSummary();
 
         return (
-            <div>
-                <p>{webhookLegend}</p>
+            <React.Fragment>
+                <p>{WEBHOOK_LEGEND}</p>
```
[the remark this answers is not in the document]
ycleptkellan: Yeah more like a description of the form's purpose. what about `WEBHOOK_DESCRIPTION`?

src/components/routers/WebhookRouter.tsx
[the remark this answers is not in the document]
ycleptkellan: http://gph.is/2Eq5ph3

ycleptkellan:
@ericnewcomer addressed your requests here as well 👍

ycleptkellan:
probably best to merge this first when both are ready. should be a few tiny conflicts but nothing significant

PR #138, opened 2018-02-14
pin deps

ycleptkellan:
- pins our dependencies at exact versions and applies the `--save-exact` flag to all NPM packages installed within the repo. 
- updates `package-lock.json`
- updates TypeScript and Enzyme to newest versions


ycleptkellan:
I'd love for us to incorporate a tool like [GreenKeeper](https://greenkeeper.io/), and in order for it to create PRs for dependency updates [we need to pin our dependencies](https://greenkeeper.io/faq#pinning-versions).

ericnewcomer:
Part of @ycleptkellan's march to use every library possible! Actually think GK looks pretty great. Of course there is still risk with updating libs even if your tests pass, but probably better to tackle that with a bit more frequency than we have.

ycleptkellan:
😂

PR #147, opened 2018-02-26
load google fonts

ycleptkellan:
figure it's better to provide these ourselves than to request them from the Google Fonts CDN. thoughts?


ericnewcomer:
I think this depends on if we are going to bundle them intelligently. Argument for a CDN is the user might already have that font cached in their browser before they even visit us. Definitely nicer from a development perspective to have them local (which we hacked with a conditional load in the old editor, don't think we want to carry that approach forward).

I think l'm fine bundling all the fonts we use. Curious if @nicpottier has an opinion there.

nicpottier:
Roboto does happen to be one where the hit rate to a CDN would be high, but ya, I'm fine optimizing for that later.

ycleptkellan:
sounds good 👍

PR #151, opened 2018-02-26
character-counter

### LINKED ISSUE #149: Character counter tweaks / help
Couple comments:
 * I don't think we need the question mark, we can just show the pop up when you go over the numbers
 * The help text should use the numbers that are currently set instead of static ones

flow2 2018-02-26 11-08-12

ycleptkellan:
resolves #149


nicpottier:
Man the netlify previews are so badass.

I think the reverse count is maybe a bit more confusing than help, how about we only count upwards and show # of estimated segments? @ericnewcomer thoughts there?

For the text then it can be something easier like:
```
Your message will be at least n characters long. We estimate it will take m segments to send over SMS.
```

ycleptkellan:
that sounds much better to me 👍

sets the right expectations for messages containing variables

nicpottier:
Ya, maybe we should use `about n characters` long actually, since substitution could make it shorter, but anyways seems a bit clearer.

ericnewcomer:
Agree, counting up is better, especially since less and less customers will be using SMS going forward and this isn't a concern over ip messaging.

On display, can we have the numbers be smaller with a bit of padding width on the `/`?

src/component/form/TextInputElement/CharCount.tsx
```diff
@@ -0,0 +1,24 @@
...
+const CharCount: React.SFC<CharCount> = ({ count, parts }) => (
+    <div className={styles.count}>
+        <div className={`${styles.tooltip}`}>
+            {count}/{parts}
+            <span className={styles.tooltipText}>
+                The character-counter provides an estimate of the number of messages your response
+                is expected to be split into by carriers. For example, the count <b>{count}/{parts}</b>
+                {' '}indicates that the current message will be split into <b>{parts}</b> messages and you have{' '}
+                <b>{count}</b> characters left until it will be split into <b data-spec="parts-next">{parts + 1}</b> messages.
+            </span>
```
ericnewcomer: Careful in your new blurb to deal with language difference for plurality, missing here.

ycleptkellan:
sounds good 👍 updated the PR LMK if there's anything I missed/should change

nicpottier:
Do we deal with the difference with GSM7 vs not? I think I remember you saying you wanted to do that Kellan? If not, that's fine, but if so perhaps we should add some verbage when they switch into UTF to say why it is more segments.

ycleptkellan:
we do yes, how do you think we should call that out? Conditionally adding a statement like `Each segment may be limited in size because your message contains special characters`? We could even display those characters like this does: 

https://message-details.now.sh/

nicpottier:
Ya, I think if we know the characters, then we can say something like:

`Your message will require more segments to send via SMS because it contains the character(s): :)`

ycleptkellan:
tooltip

What do we think about something along these lines?

ycleptkellan:
updated to include that warning and a list of the 'offending' characters, LMK what ya think. Note that we replace some characters off the bat: 

https://github.com/nyaruka/floweditor/blob/character-counter/src/component/form/TextInputElement/TextInputElement.tsx#L271

Any others we should replace?

nicpottier:
Don't want to bike shed this too much, I'm fine with this for the time being though I probably wouldn't treat those special characters quite so specially. :)

ycleptkellan:
Aye, thoughts @ericnewcomer?

ericnewcomer:
Don't really take issue to the extended character treatment. Sometimes it can be super hard to know why (see pasted quotes). I think this is a good stopping point -- we can revisit it in the future. The verbiage to be "about" this long is a little funny. Ideally I think we'd have different verbiage if there's an expression, but I think this is totally good enough to move on.

PR #152, opened 2018-02-28
Experiment: Lighter weight connectors

ericnewcomer:
Not intended for merging, just a small experimental PR for discussion

### DIFF
--- a/src/component/Exit.scss
+++ b/src/component/Exit.scss
@@ -77,14 +77,15 @@
     .endpoint {
         position: relative;
         background: $terminal_color;
-        border: 2px solid $terminal_color - #111;
-        width: 10px;
-        height: 10px;
+        border: 1px solid $terminal_color - #111;
+        width: 8px;
+        height: 8px;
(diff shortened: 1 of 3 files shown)


ycleptkellan:
love the direction

nicpottier:
neato! arrow heads are a bit hard to grab but definitely like things being lighter weight.

rowanseymour:
👍 i like that the arrows don't quite touch the node

norkans7:
will it still be possible to hover on the connection and it turns green, that helps in crazy flows to see were the arrow is going

ericnewcomer:
@nicpottier the actual grab targets shouldn't be smaller, those are independent, but can definitely tweak that.

@rowanseymour ya, I like the larger gap there too. Obviously we can move that in either direction if we want to tweak it more.

@norkans7 you are right, we do need to come up with a way to track these paths (highlighting on hover would be one way to do that), we have some other ideas to explore to make congested flows a bit easier to grok.

PR #154, opened 2018-03-01
Experiment: Curves

ericnewcomer:
Things to look at:
* Roundy nodes, modals, buttons, and endpoints
* Bit more contrast for the canvas

Compare with: https://hardcore-hypatia-8b3d76.netlify.com/

I think I like this direction, thoughts?

### DIFF
--- a/src/component/Button.scss
+++ b/src/component/Button.scss
@@ -10,6 +10,7 @@
 
   cursor: pointer;
 
+  border-radius: 4px;
   background: $button_dark;
   background-image: -webkit-linear-gradient(top, $button_light, $button_dark);
   background-image: -moz-linear-gradient(top,$button_light, $button_dark);
--- a/src/component/Exit.scss
+++ b/src/component/Exit.scss
@@ -77,14 +77,15 @@
     .endpoint {
         position: relative;
         background: $terminal_color;
-        border: 2px solid $terminal_color - #111;
-        width: 10px;
-        height: 10px;
+        border: 1px solid $terminal_color - #111;
+        width: 8px;
+        height: 8px;
         margin: 0 auto;
         top: 6px;
         -webkit-box-sizing: content-box !important;
         box-sizing: content-box !important;
-        margin-top: -14px;
+        margin-top: -12px;
+        border-radius: 4px;
     }
     &.confirm_delete {
         .activity {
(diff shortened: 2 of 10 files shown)


ycleptkellan:
Totally agree, love the rounded nodes, editor modal, buttons. 

My 2¢:

- I like the added contrast between the canvas and the nodes, but it strikes me as too gray given the current color scheme. Maybe the lines need some color...a blue perhaps?
- I think the connectors should have a tad more width. Perhaps a new blue?
- I think each node should have more box shadow in its resting state. 
- I love the tomato color of the unconnected endpoints in the current version of the editor, perhaps we should keep it? FWIW I think matte, lighter colors like tomato, cyan and the blue highlight in the v2 autocomplete menu would give the editor a more modern feel. 
- I really like the box shadow on the [Netlify auth modal](https://cdn.netlify.com/56faa9c5698de71195f9d1efb583632855ff3719/15519/img/blog/open-sauced-login.png), perhaps we should find a way to add some to the `NodeEditor` modal?

nicpottier:
Definitely like the roundies. Canvas feels too dark to me, love the roundy modals as well.

I am going to be super unhelpful in that ya I yearn for something more fun or modern but I have absolutely no idea how to achieve that. Brighter colors? Less dark? I have no idea. :(

rowanseymour:
Delicious curves 👍 . I think problem with the darker canvas is there's now less definition around the edges of the nodes because their outline is a very similar shade of gray.

ericnewcomer:
Nah, feedback is all helpful even without ideas. This is just exploring one thing (well a couple things I guess). Lots more to explore and do of course, I’m just trying to get early feedback on some ideas before polishing anything. Agree the borders wash out a bit, the question is more, are we interested in something other than white?

nicpottier:
other than white where? the background? I think a bright background is good personally, but could be convinced otherwise I'm sure. @dodobas pasted a link in slack that was a good reminder on some concepts we could work in to try to lighten things a bit as well: https://medium.com/refactoring-ui/7-practical-tips-for-cheating-at-design-40c736799886

ycleptkellan:
great find @dodobas! off the bat i can see some things that would help:

- inset shadow on inputs
- more shadow for each node
- lose background color for secondary buttons (secondary button example)

ericnewcomer:
Sounds unanimous on showing some curves. I'll work in some of these ideas for the next step and throw up another strawman.

PR #155, opened 2018-03-01
Experiment: Brighter

ericnewcomer:
Not necessarily sold on all things, but consider the following for feedback / ideas:

* brighter palette
* more spacing
* lighter connection treatment
* secondary buttons only on hover
* remove some rules
* shadow boxes on nodes and modal (not wild on the latter, feels rather bootstrappy)


nicpottier:
I like this direction, feels light.. the bottom warts for the connectors felt really weird at first but they are growing on me. Few nicks here and there but know that's not the point of these. Arrows could probably be a bit darker, might be taking it a bit too far.

rowanseymour:
😍 the new nodes outlines, warts included
😍 the fact the legacy shade of blue is gone
😕 the connectors seem a bit faint

ericnewcomer:
I like the lightness of the connected doodads since most things will be connected -- makes everything feel a bit more airy. I do worry about discoverability for click to disconnect (though admittedly it's not super obvious in the old editor either).

nicpottier:
Seems like making an X appear on rollover would help with discovering that.. true that warts are lighter weight than boxes, so do like that. Really like new colors.

nicpottier:
As an aside and I know this is all detail work to do later, but I think a lot of the feel of it being light is being super consistent with spacing everywhere, ie, grid the hell out of it. We are bit all over the place right now. (which is expected, just noting this for when we get to the polish stage of experiments)

ericnewcomer:
The more I stare at this the more I really like the light connectors (especially for really busy flows), but willing to acquiesce to the crowd. 

One thing I still want to look at is merged paths for common destinations (which is more of a jsplumb task and is a bit involved) -- but that might get us that simplified feeling in an even bigger way.

ycleptkellan:
i love this so much. digging the color scheme so far–airy as you said, pleasant to look at. i think the color of the endpoints should match that of the connectors and, if we go that route, we could apply some shadow to the half of the connector that overlaps with the exit, which would create even more depth/texture. i don't think the color of the connectors should be darkened too much but there's a little room to do so if need be. the modal shadow needs some work but i think we'd be remiss to not eventually give it some `box-shadow` love. that might even mean lightening the color of the modal overlay. 

another thought, probably unrelated: i think we can have a lot of fun with displaying stacked nodes at this border radius. what if they were staggered like a deck of cards and each one popped up a little bit on hover?

ycleptkellan:
some fun color resources:

- https://flatuicolors.com/
- https://colorwise.io/
- http://ourownthing.co.uk/gradpad.html?ref=producthunt in case we're open to gradients

nicpottier:
One other thing I find slightly distracting is that the depth of the line is different than the node and connector. As soon as we add drop shadows (which I like) I feel we need to make sure there's some logical depth to the depth of things and that's one that pops out at me.

nicpottier:
Those actions put on a few pounds!

norkans7:
I like the boxes the are so much better looking
those on other languages are broken I guess in case you did not notice

ericnewcomer:
@nicpottier did they? didn't intend to change the actions much outside of spacing and removing borders.

@norkans7 ya, localization is goobered, not looking at that just yet.

ycleptkellan:
aye, looks good 👍

PR #153, opened 2018-02-28
multi-arg-case

ycleptkellan:
attempt at resolving #128

PR #157, opened 2018-03-01
Result names

### LINKED ISSUE #123: NodeEditor loses 'Result Name' ref when form type changes

ycleptkellan:
resolves #123


ycleptkellan:
will add comments once #153 is merged. feels good to be pushing state up to `NodeEditor` as we don't need to worry about updating the state of child elements.

src/component/form/CaseElement.tsx
```diff
@@ -305,15 +306,15 @@ export default class CaseElement extends React.Component<CaseElementProps, CaseE
...
+        const invalidExitErr = `A category name is required when using "${
             this.state.operatorConfig.verboseName
         }."`;
 
         if (/between$/.test(this.state.operatorConfig.type)) {
             if (
                 !this.state.arguments.length ||
                 this.state.arguments.length !== 2 ||
-                (!this.state.arguments[0] && !this.state.arguments[0])
+                (!this.state.arguments[0].trim() || !this.state.arguments[1].trim())
```
ycleptkellan: missed this is #157

src/component/form/CaseElement.tsx
```diff
@@ -305,15 +306,15 @@ export default class CaseElement extends React.Component<CaseElementProps, CaseE
     public validate(): boolean {
         const errors: string[] = [];
         // prettier-ignore
-        const INVALID_EXIT_ERR = `A category name is required when using "${
+        const invalidExitErr = `A category name is required when using "${
```
ycleptkellan: realize this shouldn't be constant-case as it contains a dynamic value

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -89,7 +93,19 @@ export interface NodeEditorProps {
 export interface NodeEditorState {
     config: Type;
     show: boolean;
-    setResultName: boolean;
+    resultName: string;
```
ycleptkellan: moving result name to `NodeEditor` state so we can persist it across forms

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -611,6 +808,252 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
         this.props.onUpdateRouter(node, type, previousAction);
     }
 
+    private updateSwitchRouter(kases: CaseElementProps[]): void {
```
ycleptkellan: moving each router's router-update method up to `NodeEditor` since, for the most part, all the info each of these methods need exists in `NodeEditor`. the only case where that isn't true is `SwitchRouter`, but that will change if we lift `SwitchRouter`'s `cases` state up to `NodeEditor`. if/when we bring in an centralized state management library, we can extract these as 'actionCreators' in Redux parlance. i'd also like to see us move away from maintaining a widget map/extracting widget state through React's `ref` API.

src/component/routers/WebhookRouter.tsx
```diff
@@ -72,39 +52,35 @@ export const mapHeaders = (headers: Headers): Header[] =>
...
-export const getInitialState = (action: CallWebhook): WebhookState => {
+export const getInitialState = (action: CallWebhook): WebhookRouterState => {
+    let state = initialState;
     if (action.type === 'call_webhook') {
-        initialState.method = action.method;
+        state = { ...state, method: action.method };
         if (action.headers && Object.keys(action.headers).length) {
             const existingHeaders = mapHeaders(action.headers);
-            initialState.headers.unshift(...existingHeaders);
+            state = {
```
ycleptkellan: is this OK? just trying to keep state pure as we have render two `WebhookRouter` components when the user switches to that form

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -681,11 +1124,27 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
     }
 
     private getSides(): Sides {
-        const action: AnyAction = getAction(this.props, this.state.config);
+        const action = getAction(this.props, this.state.config);
```
ycleptkellan: i do see us refactoring this in the future to be more specific about which props should be passed to each form.

ycleptkellan:
@ericnewcomer this one's ready for ya!

ericnewcomer:
This looks like it might be new as I can't repro this on master:
https://monosnap.com/file/vnh2WVCe2r4wsqBzr6ifUD7FdlhZkS

src/component/routers/WebhookRouter.tsx
ericnewcomer: ya, I think this is alright.. at least the object being spread is right there above it so its pretty clear what's going on there.

ycleptkellan:
@ericnewcomer thanks for catching, addressed here: https://github.com/nyaruka/floweditor/commit/a878fda241a7c429925e9fe483dbb3cd31ec7086#diff-02c44eca42688a76ee866deddf5f19b4

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -48,17 +50,16 @@ interface Sides {
...
-    action?: AnyAction;
-    localizations?: LocalizedObject[];
     definition: FlowDefinition;
     config: Type;
     ComponentMap: ComponentMap;
     updateAction(action: AnyAction): void;
     onBindWidget(ref: any): void;
     onBindAdvancedWidget(ref: any): void;
-    updateRouter(node: Node, type: string, previousAction: AnyAction): void;
+    updateRouter: Function;
```
ericnewcomer: Can you explain this change?

src/component/NodeEditor/NodeEditor.tsx
ericnewcomer: This still leaves me a little sad to not have these bundled with the presentation, but I guess I get the reasoning here.

src/component/NodeEditor/NodeEditor.tsx
ycleptkellan: yes using the generic ‘Function’ type since the value of the prop will one of ‘updateSwitchRouter’, ‘updateSubflowRouter’, ‘updateGroupRouter’ or ‘updateWebhoomRouter’

src/component/NodeEditor/NodeEditor.tsx
ycleptkellan: Ya I def not ideal; I have some ideas about this

PR #161, opened 2018-03-02
hot load scss globals

ycleptkellan:
- only apply `ExtractTextPlugin` to preview, production builds
- load global, component-scoped styles separately
- hot reload tweaks to `global.scss` and `variables.scss` files


ycleptkellan:
@ericnewcomer this should make life easier for ya 👍

webpack/webpack.dev.js
```diff
@@ -73,6 +74,17 @@ const devConfig = {
     ],
     module: {
         rules: [
+            {
```
ycleptkellan: removing `ExtractTextPlugin` from dev config as `react-hot-load` takes care of hot loading css modules so long as the `ExtractTextPlugin` isn't applied to our css loaders.

webpack/webpack.dev.js
```diff
@@ -73,6 +74,17 @@ const devConfig = {
...
+                test: /\.s?css$/,
+                include: [paths.component],
+                use: ['style-loader', typingsForCssModulesLoader(), postCSSLoader, 'sass-loader']
+            },
+            {
```
ycleptkellan: treating global styles (`global.scss` at the root of the `/src` directory) as a global style sheet.

webpack/loaders.js
```diff
@@ -0,0 +1,22 @@
+exports.typingsForCssModulesLoader = (prod = false) => ({
```
ycleptkellan: this takes the place of the `css-loader` when loading styles from `/src/component` since we apply types to our styles.

PR #163, opened 2018-03-08
Create expiration route when calling out to a flow

ericnewcomer:
Fixes nyaruka/floweditor/issues/113

@ycleptkellan could use a hand with the proper way to test `updateSubflowRouter` to ensure it creates the correct exits. Or this presumably is easier with redux?

### DIFF
--- a/src/component/NodeEditor/NodeEditor.test.tsx
+++ b/src/component/NodeEditor/NodeEditor.test.tsx
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import CompMap from '../../services/ComponentMap';
 import NodeEditor, { NodeEditorProps, mapExits, isSwitchForm, getAction } from './index';
 import { getBaseLanguage } from '../';
 import { V4_UUID } from '../../utils';
-import { getTypeConfig, typeConfigList } from '../../config';
+import { getTypeConfig, typeConfigList, endpointsPT } from '../../config';
 import { WaitType } from '../../flowTypes';
 import { resolveExits, hasWait, hasCases, groupsToCases } from './NodeEditor';
 import { extractGroups } from '../routers/GroupRouter';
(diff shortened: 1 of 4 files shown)


src/component/NodeEditor/NodeEditor.test.tsx
```diff
@@ -224,4 +222,117 @@ describe('NodeEditor >', () => {
...
+            );
+
+            expect(wrapper.find('SubflowRouter').prop('updateRouter')).toEqual(
+                wrapper.instance().updateSubflowRouter
+            );
+
+            wrapper.instance().submit();
+
+            expect(updateSubflowRouterSpy).toHaveBeenCalled();
+            expect(updateRouterSpy).toHaveBeenCalledWith(
```
ycleptkellan: assumption here is if `ComponentMap.getDetails` doesn't return a `ComponentDetails` interface of type `subflow` but `NodeEditor`'s `updateRouter` method is called with an object containing two exits and cases, the right thing happened.

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -806,7 +806,7 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
         this.props.onUpdateAction(this.props.node, act);
     }
 
-    private updateRouter(node: Node, type: string, previousAction?: Action): void {
+    public updateRouter(node: Node, type: string, previousAction?: Action): void {
```
ycleptkellan: these should be public if we're accessing them w/ `jest.spyOn`

src/component/form/FlowElement.tsx
```diff
@@ -24,13 +24,14 @@ export default class FlowElement extends React.Component<FlowElementProps, FlowS
...
 
-        const flow: SearchResult = this.props.flow_uuid
-            ? {
-                  name: this.props.flow_name,
-                  id: this.props.flow_uuid,
-                  type: 'flow'
-              }
-            : null;
+        const flow =
+            this.props.flow_uuid && this.props.flow_name
```
ycleptkellan: noticed we weren't checking for both

ycleptkellan:
digging the canvas btw

src/component/NodeEditor/NodeEditor.test.tsx
```diff
@@ -224,4 +222,133 @@ describe('NodeEditor >', () => {
...
+                        context: { endpoints },
+                        childContextTypes: {
+                            endpoints: endpointsPT
+                        }
+                    }
+                );
+
+                wrapper.instance().submit();
+
+                expect(updateSubflowRouterSpy).toHaveBeenCalled();
```
ycleptkellan: making sure `updateSubflowRouter` is taking cases, exits from existing node

PR #165, opened 2018-03-08
POC Redux Integration

ycleptkellan:
Integrates Redux w/ the root component while including the usual suspects: reducers, actions, async actionCreators, devtools middleware (you'll need to install [this](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd) or [this](https://addons.mozilla.org/en-US/firefox/addon/remotedev/)).


ycleptkellan:
tweaks, comments forthcoming

ericnewcomer:
Ya, definitely gonna need a walkthrough. With `react-redux` I expected to see presentation components wrapped with `mapStateToProps` and `mapDispatchToProps`. And honestly with those, I have a hard time wrapping my head around the right way to do that with all the things the `FlowMutator` is currently responsible for.

ycleptkellan:
i'm working on moving FlowMutator` over right now as i found the app can't run in production (on Netlify) otherwise. i did move to the typical `react-redux` container pattern (https://github.com/nyaruka/floweditor/pull/165/files#diff-723500c5a27076b3f7ad4bd0d8a61defR627) so hopefully that makes things easier. `ComponentMap` is already in Redux, you can see those changes in the `initialState`, `actions`, and `actionCreators`. Will connect the components using the `ComponentMap` and `Mutator` once `Mutator` is gtg. Hopefully that's all that needs to be done to get things running

ycleptkellan:
making good progress here, just fine-tuning at this point

ericnewcomer:
Let me know when I should pull this down and poke through it

ycleptkellan:
OK @ericnewcomer this one's ready for ya

ycleptkellan:
i waited on the tests since we aren't really sure about redux but refactoring them shouldn't be too painful

src/component/Flow.tsx
```diff
@@ -1,366 +1,131 @@
...
+    updateCreateNodePositionA: (
+        createNodePosition: Position
+    ) => { type: Constants; payload: { createNodePosition: Position } };
 }
 
 export interface Translations {
     [uuid: string]: any;
 }
 
-const REPAINT_DURATION = 600;
```
ycleptkellan: this is the application of the `Config` component I was talking about, it's actually pretty similar to the `Consumer` component React's new context API provides. none of those ugly errors for components that consume context in this way

src/component/Flow.tsx
```diff
@@ -385,24 +146,31 @@ export default class Flow extends React.Component<FlowProps, FlowState> {
...
+                const dragPoint = pendingConnection;
+                this.Plumber.recalculate(ghostNode.uuid);
+                this.Plumber.connect(dragPoint.exitUUID, ghostNode.uuid);
 
-                // save our position for later
+                // Save our position for later
                 const { left, top } = snapToGrid(
-                    this.ghost.ele.offsetLeft,
-                    this.ghost.ele.offsetTop
+                    this.ghost.wrappedInstance.ele.offsetLeft,
```
ycleptkellan: `react-redux`'s `connect` method produces a new component that keeps a reference to the component it wraps in the `wrappedInstance` key. that's what we have to reference if we pass a ref to a connected component.

src/component/form/TextInputElement/TextInputElement.tsx
```diff
@@ -478,3 +474,11 @@ export default class TextInputElement extends React.Component<TextInputProps, Te
         );
     }
 }
+
+const mapStateToProps = ({ typeConfig, resultNames }: ReduxState) => ({ typeConfig, resultNames });
+
+const ConnectedTextInputElement = connect(mapStateToProps, null, null, { withRef: true })(
```
ycleptkellan: we use the `withRef` property to indicate to `connect` that we're passing a ref to the wrapped component

src/redux/actionTypes.ts
```diff
@@ -0,0 +1,314 @@
+import Constants from './constants';
```
ycleptkellan: declaring types for our actions

src/redux/actions.ts
```diff
@@ -0,0 +1,233 @@
+import { Language } from '../component/LanguageSelector';
```
ycleptkellan: i followed the convention of creating low-level actions that are then dispatched by the action creators in the `actionCreators.ts`

src/redux/initialState.ts
```diff
@@ -0,0 +1,110 @@
...
+export interface CompletionOption {
+    name: string;
+    description: string;
+}
+
+export interface PendingConnections {
+    [uuid: string]: DragPoint;
+}
+
+export interface ReduxState {
```
ycleptkellan: the component's initial state; includes some properties on the `ComponentMap` and `FlowMutator` singletons

src/redux/reducers/components.ts
```diff
@@ -0,0 +1,13 @@
+import initialState, { ReduxState } from '../initialState';
```
ycleptkellan: each of these reducers is concerned w/ only one key in our store and we compose them in the root of this directory w/ the `combineReducers` helper

src/redux/actionCreators.ts
```diff
@@ -0,0 +1,1474 @@
+import * as isEqual from 'fast-deep-equal';
```
ycleptkellan: herein lies everything we were previously doing in `ComponentMap` and `FlowMutator`. the `redux-thunk` middleware passes us `redux`'s `getState` method, which allows us to read from state at any point. action creators like `addNode` update our flow definition so we need to call `getState` right after to get that updated definition if we want to read from it right after we add a node (some examples of that below). i imagine we can optimize these for purity as we go (probably not great to be calling `getState` everywhere).

src/redux/actionCreators.ts
```diff
@@ -0,0 +1,1474 @@
...
+export const setFreshestNode = (freshestNode: Node) => (
+    dispatch: DispatchWithState,
+    getState: GetState
+) => {
+    if (!isEqual(freshestNode, getState().freshestNode)) {
+        dispatch(updateFreshestNode(freshestNode));
+    }
+};
+
+export const setUserClickingAction = (userClickingAction: boolean) => (
```
ycleptkellan: added these 'buffer' methods so we aren't updating state unless we need to. makes for a cleaner  devtools log but not super adamant about this in case there are any objections

src/redux/actionCreators.ts
```diff
@@ -0,0 +1,1474 @@
...
+                {
+                    name: `run.results.${resultNameKey}.category`,
+                    description: `Category for "${resultNames[resultNameKey]}"`
+                }
+            );
+        }
+        return resultNameList;
+    }, []);
+
+    dispatch(setContactFields(existingFields));
```
ycleptkellan: this is all synchronous save the action creators that use the timeout API

src/redux/actionCreators.ts
```diff
@@ -0,0 +1,1474 @@
...
+            ({ operand } = getState().nodeToEdit.router as SwitchRouter);
+        }
+    }
+
+    dispatch(updateResultName(resultName));
+    dispatch(setShowResultName(resultName.length > 0));
+    dispatch(updateOperand(operand));
+    dispatch(setNodeDragging(false));
+    dispatch(setNodeEditorOpen(true));
+};
```
ycleptkellan: this file is way too large, we should talk about how we want to organize them

src/redux/configureStore.ts
```diff
@@ -0,0 +1,25 @@
+import { applyMiddleware, createStore } from 'redux';
+import { composeWithDevTools } from 'redux-devtools-extension';
+import thunk from 'redux-thunk';
```
ycleptkellan: fun fact: `redux-thunk` is super tiny https://github.com/gaearon/redux-thunk/blob/master/src/index.js

PR #166, opened 2018-03-09
icier canvas, a lot less grey

ericnewcomer:
* fixes some detail issues and border intersections
* lighter still

### DIFF
--- a/src/component/Button.scss
+++ b/src/component/Button.scss
@@ -49,7 +49,7 @@
     border: 1px solid transparent;
 
     &:hover {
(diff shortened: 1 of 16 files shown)


nicpottier:
Love the direction on this stuff, not positive about the blue drop shadow though, feels a bit glowy to me.

ericnewcomer:
Toned down the shadow, the still blue-based. This pr also focus on lots of details and parameterizes the things like the shadow stuff so it's easy to experiment and dial into something we all like.

Issue #114, opened 2018-01-09
Auto expanding canvas


ericnewcomer:
Current the editing canvas is set arbitrarily large, need to make it only grow as needed.

PR #167, opened 2018-03-10
Canvas updates

ericnewcomer:
* Allow parent container to determine canvas size
* Honor 20 x 20 grid for movement and node placement

Fixes: #114

### DIFF
--- a/preview/src/index.html
+++ b/preview/src/index.html
@@ -2,9 +2,20 @@
 <html>
 
 <head>
-    <meta charset="utf-8">
     <title>Flow2</title>
+    <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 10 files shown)


preview/src/index.html
```diff
@@ -2,9 +2,20 @@
 <html>
 
 <head>
-    <meta charset="utf-8">
     <title>Flow2</title>
+    <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <style>
+        html, body {
```
ycleptkellan: Would it make sense to move this to the root component, perhaps by wrapping the `#editor-container` `div` in another `div`?

ycleptkellan: (that outer `div` being the `#flow-editor` `div`)

PR #170, opened 2018-03-12
Smooth drag

ericnewcomer:
Snap to grid after node placement instead of on mouse move.


ericnewcomer:
Actually doesn't move too much after placement to offend me, I think I'm cool with this. Thoughts?

PR #171, opened 2018-03-14
contain styles

ycleptkellan:
want to make sure we're scoping styles to the component so long as we can help it

### DIFF
--- a/preview/src/index.html
+++ b/preview/src/index.html
@@ -5,17 +5,6 @@
     <title>Flow2</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <style>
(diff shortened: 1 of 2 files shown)


ericnewcomer:
Not sure I understand here.. Isn't preview effectively a mock of an external site including / using the component?

ycleptkellan:
sorry, perhaps i'm not clear on the intent there. are the styles under the `#flow-editor` selector simulating something a consumer would apply?

ericnewcomer:
Yup, that is my understanding. Isn't `#flow-editor` the name of the element in the preview that's embedding our component?

ycleptkellan:
hah wow i'm a DA, totally missed that. closing

PR #172, opened 2018-03-14
fix confirm delete style for router nodes

ycleptkellan:
gives border-radius to title bar when user is confirming node deletion


src/component/TitleBar/TitleBar.scss
```diff
@@ -16,18 +16,20 @@
         position: absolute;
         margin-top: -27px;
         border: 0px solid #c14c46;
+        border-top-left-radius: 7px;
+        border-top-right-radius: 7px;
```
ericnewcomer: This needs to be conditional since you are breaking title bars for any action after the first one that is being removed.

ycleptkellan:
doh, thanks i think i have that addressed now. LMK if there's a better way to handle that styling, not super confident w/ CSS

src/component/TitleBar/TitleBar.tsx
```diff
@@ -1,11 +1,13 @@
...
+import * as classNames from 'classnames/bind';
+import { Node } from '../../flowTypes';
 import * as styles from './TitleBar.scss';
 
 interface TitleBarProps {
     title: string;
+    node: Node;
     onRemoval(event: React.MouseEvent<HTMLDivElement>): any;
-    className?: string;
+    __className?: string;
```
ericnewcomer: why move to dunder? what's the convention?

ycleptkellan: we've sort of already set that convention elsewhere, the idea being that it could be confused w/ React's `className` property. borrowed it from EventBrite's style guide: https://github.com/eventbrite/javascript/tree/master/react#classname-prop

ericnewcomer:
I think a css only solution should be possible and we should aim for that where it simplifies things. Maybe styling all with curves and then have a style to square off the extra actions with something like..

```css
not(:first-child) {
  border-top-left-radius: none;
  border-top-right-radius: none;
}
```

src/component/TitleBar/TitleBar.tsx
ericnewcomer: I'm fine with it, just want to understand so we can aim for consistency. Maybe we should draw up our own conventions.

ycleptkellan: ya that would be awesome, if we need inspiration the airbnb style guide is pretty popular.

ycleptkellan: will take a look and see if there's anything else we're already doing so we can  discuss 👍

ericnewcomer: I'm fine adopting an existing one and making any amendments from there if we feel strongly enough to deviate. No reason to waste a ton of time on it, just having an understanding is the useful bit.

ycleptkellan: ok sounds good, if like airbnb's style i can add a config to tslint. also, in case you haven't already installed it, the tslint vscode plugin will show you linting errors so that might be helpful

ericnewcomer: Yup, in a separate PR.

ycleptkellan:
totally agree, super clean

ycleptkellan:
having trouble w/ that approach, it looks like the div w/ the `.remove_confirm` class is always going to be the first child...same w/ the `.titlebar` class. is there anything else we could key on? not sure if the scope of that file would allow us to select the right element

ericnewcomer:
It's only the action case we need to be worried about, so nesting under `.actions` ought to work I would think. Let me know if you want me to look closer at this, it's possible my hunch is wrong.

ycleptkellan:
ya some help would be awesome, not sure this can be done w/o conditionally applying a class like i had done originally. total SASS noob though, so i could be overlooking something

ericnewcomer:
k, move on to something else, I'll look at this next.

PR #174, opened 2018-03-14
Update actions

ycleptkellan:
update actions per https://github.com/nyaruka/goflow/issues/172


ycleptkellan:
😬 i already deployed the updated flows to the asset server so the playground won't work until this is merged

PR #180, opened 2018-03-15
show result name

### LINKED ISSUE #179: Deleting existing result name hides field
Repro:
* Open node with a result name
* Focus at the end of the name
* Delete characters one by one to give a new name (or select all and delete)
* Result Name field vanishes

Expected:
Result Name field should never vanish while the user has focus on it

ycleptkellan:
fixes #179

### DIFF
--- a/src/component/NodeEditor/NodeEditor.test.tsx
+++ b/src/component/NodeEditor/NodeEditor.test.tsx
@@ -3,7 +3,7 @@ import { shallow, mount } from 'enzyme';
 import CompMap from '../../services/ComponentMap';
 import NodeEditor, { NodeEditorProps, mapExits, isSwitchForm, getAction } from './index';
 import { getBaseLanguage } from '../';
-import { V4_UUID } from '../../utils';
+import { V4_UUID, getSpecWrapper } from '../../utils';
 import { getTypeConfig, typeConfigList, endpointsPT } from '../../config';
 import { WaitType } from '../../flowTypes';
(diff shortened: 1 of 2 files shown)


ericnewcomer:
How hard is it to add a test for this?

ycleptkellan:
not so hard, happy to add

ycleptkellan:
ok added a few tests for result name display

src/component/NodeEditor/NodeEditor.test.tsx
```diff
@@ -221,6 +224,48 @@ describe('NodeEditor >', () => {
...
+                />
+            );
+
+            expect(getSpecWrapper(wrapper, 'name-field').prop('value')).toBe(
+                waitNode.router.result_name
+            );
+
+            wrapper.setState({ resultName: '' });
+
+            expect(getSpecWrapper(wrapper, 'name-field').prop('value')).toBe('');
```
ericnewcomer: is this ensuring that it is visible?

ycleptkellan: ya the result of the `getSpecWrapper` call is the `TextInputElement` so if this doesn't throw it's both visible and it contains an empty string

ericnewcomer: Ok, that’s helpful. Did you confirm the test failure without your fix?

ycleptkellan:
@ericnewcomer updated test to be clearer, fail for previous case

Issue #156, opened 2018-03-01
Make people opt out of saving results


nicpottier:
I think when we were first redoing the editor we were kinda in this mode that results are heavy to save for flows and we should make the user explicitly tell us they want to save. I think that's going to be a bit of a disaster if we ship that because I can just predict that the majority of people are going to space about saving a result and therefore lose data.

So I think at least for any ruleset that is a "wait" we need to default to saving so that they don't lose data that they realize only later they don't have and can no longer recover.

Note that our schema has changed since then too, so we care a lot less about saving results than we did, so having a bias towards keeping data is good. (though I think from a convenience point of view not requiring names for non-waits still probably makes sense)

ericnewcomer:
Ya, I think that totally makes sense.

Should we require a name or just provide one for them like we do in the old world? ie, "Response 5", etc. I think the pre-filled response that they can edit served us well and is worth keeping. thoughts?

PR #181, opened 2018-03-15
Make result name suggestion for new msg waits

ericnewcomer:
Fixes: #156 (this still needs unit tests)

### DIFF
--- a/src/component/Flow.test.tsx
+++ b/src/component/Flow.test.tsx
@@ -40,5 +40,16 @@ describe('Flow >', () => {
             expect(getSpecWrapper(wrapper, 'nodes').hasClass('node_list')).toBeTruthy();
             expect(wrapper.find('NodeComp')).toBeTruthy();
         });
+
+        it('should suggest appropriate response names', () => {
+            const wrapper = shallow(<Flow {...props} />, {
+                context: {
+                    endpoints
+                }
+            });
+
+            const flow = wrapper.instance() as Flow;
+            expect(flow.getSuggestedResultName()).toEqual('Response 2');
+        });
     });
 });
--- a/src/component/Flow.tsx
+++ b/src/component/Flow.tsx
@@ -2,7 +2,16 @@ import * as React from 'react';
 import * as FlipMove from 'react-flip-move';
 import update from 'immutability-helper';
 import { v4 as generateUUID } from 'uuid';
-import { FlowDefinition, Action, Position, SendMsg, Node, UINode, Dimensions, WaitType } from '../flowTypes';
+import {
+    FlowDefinition,
+    Action,
+    Position,
+    SendMsg,
+    Node,
+    UINode,
+    Dimensions,
+    WaitType
+} from '../flowTypes';
 import ComponentMap from '../services/ComponentMap';
 import NodeComp, { DragPoint } from './Node';
 import FlowMutator from '../services/FlowMutator';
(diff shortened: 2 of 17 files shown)


tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "allowJs": true,
     "watch": true,
     "alwaysStrict": true,
-    "lib": ["es6", "dom"]
+    "lib": ["es2017", "dom"]
```
ycleptkellan: why the change here?

ericnewcomer: `Object.values()` .. so certainly not necessary, but don't see any reason to avoid it. 

More here: https://github.com/Microsoft/TypeScript/issues/10768

ericnewcomer: Hmmm.. but that is failing the node 6 build, so maybe there's one.

ycleptkellan: nice, can't wait for that to land

ycleptkellan: `Object.entries()`?

src/component/Flow.tsx
```diff
@@ -330,9 +340,30 @@ export default class Flow extends React.Component<FlowProps, FlowState> {
 
             ghost.actions.push(replyAction);
         } else {
+            // coun't up all our msg waits
+            let responseNumber = 1;
```
ycleptkellan: should these be broken-out into helper methods, e.g. `getResponseNumber` and `getResultName`? would that make testing easier?

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -764,11 +765,9 @@ export default class NodeEditor extends React.PureComponent<NodeEditorProps, Nod
     }
 
     public open(): void {
-        const config = getTypeConfig(this.determineConfigType(this.props));
-
         this.setState({
-            show: true,
-            config
+            ...this.getInitialState(),
```
ycleptkellan: yes eric, yes...embrace the spread operator...

ericnewcomer: special case!

ericnewcomer:
@ycleptkellan what say we merge this so we don't have anything outstanding against master

ycleptkellan:
aye, merge away

src/component/Flow.tsx
```diff
@@ -294,6 +304,30 @@ export default class Flow extends React.Component<FlowProps, FlowState> {
         this.resetState();
     }
 
+    /**
+     * Gets a suggested result name based on the current number of waits
+     * in the current definition
+     */
+    public getSuggestedResultName(): string {
```
ycleptkellan: @ericnewcomer what think you about–in the future–shying away from adding methods to component classes that don't either handle a callback that changes state somewhere in the render tree or determine whether or not JSX is rendered? my reasoning is that a method that doesn't read from the component's `this` context is much easier to test (no need to render `Flow`, access it's instance and then test `getSuggestedResultName`). what do you think?

src/component/Flow.test.tsx
```diff
@@ -40,5 +40,16 @@ describe('Flow >', () => {
             expect(getSpecWrapper(wrapper, 'nodes').hasClass('node_list')).toBeTruthy();
             expect(wrapper.find('NodeComp')).toBeTruthy();
         });
+
+        it('should suggest appropriate response names', () => {
+            const wrapper = shallow(<Flow {...props} />, {
```
ycleptkellan: Alt Text

src/component/Flow.tsx
ericnewcomer: Agreed.

PR #189, opened 2018-03-17
updateRouter cleanup

ericnewcomer:


src/redux/actionCreators.ts
```diff
@@ -174,6 +177,17 @@ export const setNodeEditorOpen = (nodeEditorOpen: boolean) => (
     }
 };
 
+export const applyUpdateSpec = (updateSpec: any) => (
```
ycleptkellan: i think an `undefined` comparison might be more appropriate here, but also wondering if it isn't better for us to avoid `undefined/null/false`  comparisons in JavaScript where possible (given all the weirdness that follows). how do you feel about using a default parameter as a way to avoid that?

```
export const applyUpdateSpec = (updateSpec: any = {}) => (
	dispatch: DispatchWithState,
	getState: GetState
) => {
	if (Object.keys(updateSpec).length > 0) {
		const updatedDefinition = update(getState().definition, updateSpec);
		dispatch(updateDefinition(updatedDefinition)); 
	}
};
```

src/redux/actionCreators.ts
```diff
@@ -1078,57 +1085,31 @@ export const updateRouter = (
...
-        // We're updating
-        ({ definition: newDef } = getState());
-        const nodeDetails = getDetails(newNode.uuid, components);
-        newDef = update(newDef, {
-            nodes: { [nodeDetails.nodeIdx]: { $set: newNode } }
-        });
-        newNode = newDef.nodes[nodeDetails.nodeIdx];
-        dispatch(updateNodeUI(newNode.uuid, { $merge: { type } }));
-        dispatch(setDefinition(newDef));
+        // Dragging from somewhere means we are a new node
```
ycleptkellan: we could even make this more atomic/declarative if need be, e.g.

```
export const updateRouter = (
    node: Node,
    type: string,
    draggedFrom: DragPoint = null,
    newPosition: Position = null,
    previousAction: Action = null
) => (dispatch: DispatchWithState, getState: GetState) => {
    const { definition, components } = getState();

    if (isActionSet(node.uuid, components, definition)) {
        dispatch(routerFromAction(node, type, previousAction));
    } else {
        // Dragging from somewhere means we are a new node
        if (draggedFrom) {
            dispatch(addRouter(node, draggedFrom.exitUUID));
        } else {
            // Otherwise we are updating an existing node
            dispatch(
                applyUpdateSpec(prepSetNode(getDetails(node.uuid, components).nodeIdx, node, type))
            );
        }
    }

    console.timeEnd('updateRouter');
};
```

...`routerFromNode` and `addRouter` comprising the code in the `if (isAction(...` and `if (draggedfrom...` blocks

ericnewcomer:
@ycleptkellan updated this to remove some actions I'm thinking we don't need, also fixes #193

ycleptkellan:
nice, looks good 👍

we probably don't need any of the `set...` ACs that first check whether the argument and current state are equal

PR #196, opened 2018-03-19
fix router connection issue

ycleptkellan:
was unnecessarily resetting `actionToEdit`, `nodeToEdit` state

### DIFF
--- a/src/redux/actionCreators.ts
+++ b/src/redux/actionCreators.ts
@@ -1322,8 +1322,6 @@ export const onUpdateRouter = (
     if (nodeUUID !== newNodeUUID) {
         repaintForDuration();
     }
-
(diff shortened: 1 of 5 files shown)


src/redux/actionCreators.ts
```diff
@@ -1160,14 +1160,6 @@ export const resetNodeEditingState = () => (dispatch: DispatchWithState, getStat
...
         dispatch(updateCreateNodePosition(null));
     }
-
-    if (actionToEdit) {
-        dispatch(updateActionToEdit(null));
-    }
-
-    if (nodeToEdit) {
-        dispatch(updateNodeToEdit(null));
-    }
```
ericnewcomer: Can you explain why resetting node editing state wouldn't be nulling these?

PR #198, opened 2018-03-19
action click fix

ycleptkellan:


src/redux/actionCreators.ts
```diff
@@ -1325,20 +1325,18 @@ export const onOpenNodeEditor = (node: Node, action: AnyAction, languages: Langu
     }
 
     if (node.actions && node.actions.length) {
-        // Account for hybrids or clicking on the empty exit table
-        dispatch(setActionToEdit(node.actions[node.actions.length - 1]));
+        dispatch(setActionToEdit(action));
```
ericnewcomer: seems this should be a conditional on if the passed in action is null or not. If it is null, then dispatching against `[actions.length - 1]` would honor the other case as well.

ycleptkellan:
@ericnewcomer added an undefined-action check to both `onOpenNodeEditor` and `determineConfigType`

Issue #195, opened 2018-03-19
Connecting a split to another split blows up


ericnewcomer:
This will get into what looks like some kind of dispatch loop.

ycleptkellan:
this looks to be specific to `GroupsRouter`; `this.props.nodeToEdit` is `null` for some reason...looking into it

ericnewcomer:
I'm seeing it with non group splits: http://take.ms/HsKe7

```
floweditor.js:1 Uncaught TypeError: Cannot read property 'offsetLeft' of null
    at e.jsPlumbInstance.getOffset (VM296 floweditor.js:1)
    at Object.updateOffsets (VM296 floweditor.js:1)
    at VM296 floweditor.js:1
    at pe (VM296 floweditor.js:1)
    at e.jsPlumbInstance.revalidate (VM296 floweditor.js:1)
    at VM296 floweditor.js:1```

ericnewcomer:
Fixed in #199

Issue #190, opened 2018-03-19
Dropping connection on split opens editor


ericnewcomer:
Drag any connection onto a split node and the node editor pops up instead of just connecting the nodes.

ycleptkellan:
this only happens for me when dragging from an unconnected exit

ericnewcomer:
Fixed in #199

PR #199, opened 2018-03-19
Create a custom click handler based on mouseDown and mouseUp

ericnewcomer:
Fixes: #190, #195 

This is to work around allowing jsPlumb to manage dragging. We need to be able to detect if a node is dragging and as such the onClick event is not sufficient since it is fired after the conclusion of the the dragging action. We work around this using a mouseDown and mouseUp listener.


src/utils/index.ts
```diff
@@ -165,3 +165,29 @@ export const getLocalizedObject = (localizations: LocalizedObject[]) => {
...
+ * Creates a simple click handler via onMouseDown and onMouseUp.
+ * This is a necessity in order to let jsPlumb manage our element dragging.
+ * @param onClick
+ */
+export const createClickHandler = (
+    onClick: (event: React.MouseEvent<HTMLDivElement>) => void
+): ClickHandler => {
+    return {
+        onMouseDown: (event: React.MouseEvent<HTMLDivElement>) => {
+            this._clicked = true;
```
ycleptkellan: why not push `_clicked` to component state?

ericnewcomer: This would require any component that uses this to have `clicked` as part of its state. Not sure that's worth it since this lives only in the instance between mouse down and mouse up.

ericnewcomer:
Found one more place on `Exit` to leverage the new click handler doodad. Turns out this fixes #195 too.

PR #206, opened 2018-03-20
fix router update

ycleptkellan:


ycleptkellan:
@ericnewcomer noticing we run into issues when dragging back to a target node the source node is already connected to and then saving

ycleptkellan:
https://monosnap.com/file/cHd07OqDSdXJX02vWpjN5iMbeyZ5gt

ycleptkellan:
wondering if this fix suffices; imagine things will change when you refactor the ACs

PR #209, opened 2018-03-20
Fix reflow

ericnewcomer:
Willing to do this over again if it conflicts too hard with reducer reorg


ycleptkellan:
awesome 🙏, looks great. i added this to #210 (`thunks.ts`), lmk if i missed anything

ericnewcomer:
Looks like you got it

Issue #208, opened 2018-03-20
Reflow doesn't work


ericnewcomer:
Moving nodes on top of the first one works, but nothing past that.

Currently we rely on nodes being sorted for a couple different reasons. One, is that the flow engine assumes the starting node is the first node. The other is collision detection during reflow is optimized for an ordered list of nodes. One thought is it'd potentially be nice if we didn't re-sort nodes frequently as it invalidates the components map which is based on node indices. 

Some things to ponder:
* Only sort on demand when it's genuinely needed (saving flow def)?
* Update component map indices instead of regenerating it after a sort?
* Keep a sorted list of node uuids in state and use that where needed but defer updating the flow definition?

PR #210, opened 2018-03-20
colocate redux actions, reducers

ycleptkellan:
@ericnewcomer might be easiest to pull it down and observe the file structure. i don't think it makes sense to colocate our thunks w/ our actions/reducers _right now_ since they're all over the place as far as the actions they're dispatching but that might be a consideration for your thunk refactor. lmk if you have any suggestions for further organizing what would be the `store` directory.


ycleptkellan:
resolves #210, #208

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -385,6 +386,12 @@ export class NodeEditor extends React.PureComponent<NodeEditorProps> {
     constructor(props: NodeEditorProps) {
         super(props);
 
+        const showResultName = props.resultName.length > 0;
+
+        this.state = {
+            showResultName
```
ericnewcomer: thought you were going to keep this in the redux store?

ycleptkellan: oh sorry i thought you had suggested keeping that in `NodeEditor`, will move it back

ericnewcomer: I think since we are so close, just sticking to the policy of all state in redux for now (but scoped) makes sense. We can revisit this later, but I think this just reduces the cognitive load for someone coming into the project.

PR #213, opened 2018-03-21
initial post-redux tests

ycleptkellan:


src/store/helpers.ts
```diff
@@ -280,3 +282,37 @@ export const getCollisions = (nodes: Node[], ui: UIMetaData, nodeSpacing: number
 
     return updatedNodes;
 };
+
+export const getGhostNode = (fromNode: Node, fromNodeUI: UINode, definition: FlowDefinition) => {
```
ycleptkellan: added a helper for `ghostNode` composition so that we can generate ghost nodes in tests

src/store/nodeEditor.ts
```diff
@@ -21,12 +21,14 @@ export interface NodeEditor {
...
 
+const DEFAULT_OPERAND = '@input';
+
 // Initial state
 export const initialState: NodeEditor = {
     typeConfig: null,
     resultName: '',
     showResultName: false,
-    operand: '',
+    operand: DEFAULT_OPERAND,
```
ycleptkellan: figure `operand` initial state should be `@input` rather than `''`

src/store/flowContext.test.ts
```diff
@@ -0,0 +1,225 @@
+import Localization from '../services/Localization';
```
ycleptkellan: realizing now that i could have broken the reducers in `flowContext.ts`, `flowEditor.ts` and `nodeEditor.ts` into smaller reducers that are only concerned with a single key, e.g. `operand` in `nodeEditor.ts`, and then composed them w/ `combineReducers` (like the guy in that video did). do you think it makes sense to do so?

ycleptkellan:
tried to follow the redux testing guide w/ these as well 👍

https://redux.js.org/recipes/writing-tests

src/component/actions/CallWebhook/CallWebhook.tsx
```diff
@@ -1,8 +1,6 @@
 import * as React from 'react';
 import { CallWebhook } from '../../../flowTypes';
 
-const CallWebhookComp: React.SFC<CallWebhook> = ({ url }): JSX.Element => (
-    <React.Fragment>{url}</React.Fragment>
-);
+const CallWebhookComp: React.SFC<CallWebhook> = ({ url }): JSX.Element => <div>{url}</div>;
```
ycleptkellan: enzyme doesn't have great support for fragments yet

src/utils/index.tsx
```diff
@@ -191,3 +183,23 @@ export const createClickHandler = (
         }
     };
 };
+
```
ycleptkellan: wanted to try out a helper to make component initialization in tests a little nicer/discourage prop-spreading; what do you think?

src/component/actions/SetRunResult/SetRunResult.test.tsx
```diff
@@ -1,26 +1,24 @@
...
     });
 
-    it("should render SetRunResult with 'clear...' div when value prop isn't passed", () => {
-        const SaveFlowResultDiv = shallow(
-            <SetRunResult {...{ ...setRunResultAction, value: '' }} />
-        );
-
-        expect(SaveFlowResultDiv.text()).toBe(`Clear value for ${result_name}`);
+    it("should render SetRunResultComp with 'clear...' div when value prop isn't passed", () => {
+        const { wrapper, props: { result_name } } = setup({ value: '' });
```
ycleptkellan: overriding the 'value' prop here, for example

src/store/actionTypes.ts
```diff
@@ -121,145 +127,121 @@ interface UpdateShowResultNameActionPayload {
...
-export interface UpdateDragGroupAction {
-    type: Constants.UPDATE_DRAG_GROUP;
-    payload: UpdateDragGroupPayload;
-}
-
-export interface UpdateTypeConfigAction {
-    type: Constants.UPDATE_TYPE_CONFIG;
-    payload: UpdateTypeConfigPayload;
-}
+export type UpdateTranslatingAction = DuxAction<Constants.UPDATE_TRANSLATING, TranslatingPayload>;
```
ycleptkellan: generics are blowing my mind right now; can see a ton of uses for these elsewhere in the repo...action generic, router generic, etc.

src/component/routers/GroupsRouter.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+            it('should return false if given NOde does not have a groups router', () => {
+                expect(hasGroupsRouter(sendMsgNode)).toBeFalsy();
+            });
+        });
+    });
+
+    describe('render', () => {
+        it('should render self, children', () => {
+            const { wrapper, props: { localGroups, nodeToEdit }, context: { endpoints } } = setup();
+            expect(getResultNameFieldMock).toHaveBeenCalledTimes(1);
```
ycleptkellan: thinking we should make it a rule to callout the number of times a mock function is expected to be called, in the spirit of being super strict w/ tests, thoughts?

src/component/routers/GroupsRouter.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+            expect(wrapper.find('.instructions').exists()).toBeTruthy();
+            expect(wrapper.find('p').text()).toBe(GROUP_LABEL);
+            expect(wrapper.find('GroupsElement').props()).toEqual({
```
ycleptkellan: also, in that spirit, i think we should make it a rule to not use jest's `expect.objectContaining` assertion, which allows you to check for the presence of a subset of properties within an object. `placeholder` and `searchPromptText` are default props declared in `GroupsElement`, so i could have used it here to avoid bringing those constants in.

src/utils/index.tsx
ycleptkellan: needs a better name, that's for sure

__mocks__/axios.ts
```diff
@@ -0,0 +1,57 @@
+const axios = require.requireActual('axios');
```
ycleptkellan: mocking axios for tests

src/testUtils/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+        wrapper,
+        props,
+        context
+    };
+};
+
+/**
+ * Wait for promises in queue to resolve
+ */
+export const yieldToPromises = () => new Promise(resolve => setImmediate(resolve));
```
ycleptkellan: helper to call before making assertions on functionality that depends on async work, e.g. 

```
const { wrapper } = setup();

// assuming a click event initiates a network request
getSpecWrapper(wrapper, 'load-content').simulate('click');

// wait for promise to resolve
await yieldToPromises();

// force a re-render
wrapper.update();

expect(getSpecWrapper(wrapper, 'loaded-content')).toBe(...);

```

__mocks__/axios.ts
ycleptkellan: [this](https://blog.kentcdodds.com/but-really-what-is-a-javascript-mock-10d060966f7d) was helpful to me, new to the idea of mocking deps. previously had us using [nock](https://github.com/node-nock/nock) to intercept network requests and mock responses.

ycleptkellan: found [this](https://medium.com/@rickhanlonii/understanding-jest-mocks-f0046c68e53c) helpful as well

ycleptkellan:
@ericnewcomer ready to discuss this one, feel like i have a good POC for common cases; wouldn't want to add more tests until we're on the same page 👍

package.json
```diff
@@ -19,10 +19,10 @@
     "build:dev": "webpack --config webpack/webpack.config.js --env.NODE_ENV=dev",
     "build:prev": "webpack --config webpack/webpack.config.js --env.NODE_ENV=prev",
     "build:prod": "webpack --config webpack/webpack.config.js --env.NODE_ENV=prod",
-    "bundleStats:dev": "webpack --env.NODE_ENV=dev --profile --json > stats.json",
-    "bundleStats:dev:windows": "webpack --env.NODE_ENV=dev --profile --json | Out-file 'stats.json' -Encoding OEM",
-    "bundleStats:prod": "webpack --env.NODE_ENV=prod --profile --json > stats.json",
-    "bundleStats:prod:windows": "webpack --env.NODE_ENV=prod --profile --json | Out-file 'stats.json' -Encoding OEM",
+    "bundleStats:dev": "webpack --config webpack/webpack.config.js --env.NODE_ENV=dev --profile --json > stats.json",
```
ycleptkellan: should have made this change when webpack configs were moved to `<root>/webpack`

src/component/SelectSearch.test.tsx
```diff
@@ -1,39 +1,111 @@
...
+    resultType: 'groups',
+    placeholder: GROUP_PLACEHOLDER,
+    searchPromptText: GROUP_NOT_FOUND
+};
+
+const setup = composeSetup<SelectSearchProps>(baseProps, null, SelectSearch);
+
+const mapRespToSearchOptions = (resp: Resp) => resp.results.map(resultsToSearchOptions);
+
+const COMPONENT_TO_TEST = SelectSearch.name;
```
ycleptkellan: perhaps unnecessary? but what if the name of the component changes?

src/component/SelectSearch.test.tsx
```diff
@@ -1,39 +1,111 @@
...
+    describe('render', () => {
+        it('should render self, children with required props', async () => {
+            const selectRefSpy = jest.spyOn(SelectSearch.prototype, 'selectRef');
+            const loadOptionsSpy = jest.spyOn(SelectSearch.prototype, 'loadOptions');
+            const searchSpy = jest.spyOn(SelectSearch.prototype, 'search');
+            const { wrapper, props: { name, placeholder, searchPromptText, url } } = setup();
+            const SelectSearchInstance = wrapper.instance();
+            const asyncSelect = wrapper.find('Async');
+
+            await yieldToPromises();
```
ycleptkellan: yielding here because `SelectSearch.search` is called when `axios.get` in `SelectSearch.loadOptions` resolves

src/component/actions/ChangeGroups/RemoveGroupsForm.tsx
```diff
@@ -102,7 +102,7 @@ export class RemoveGroupsForm extends React.PureComponent<
             groupEl = (
                 <GroupsElement
                     ref={this.props.onBindWidget}
-                    name="Group"
+                    name="Groups"
```
ycleptkellan: think this makes more sense as `GroupElement` became `GroupsElement`. might make sense to do the same for `FieldElement`, `FlowsElement`?

src/component/routers/GroupsRouter.tsx
```diff
@@ -27,19 +27,20 @@ export type GroupsRouterProps = GroupsRouterStoreProps & GroupsRouterPassedProps
...
+        }
         return { name: resultName, id: kase.arguments[0] };
     });
 
-export const groupSplitExistsAtNode = (node: Node) =>
+export const hasGroupsRouter = (node: Node) =>
     hasSwitchRouter(node) && hasWait(node, WaitType.group);
 
-export class GroupsRouter extends React.PureComponent<GroupsRouterProps> {
+export class GroupsRouter extends React.Component<GroupsRouterProps> {
```
ycleptkellan: no sense making this a `PureComponent` instance since it's being wrapped by `connect`

src/component/routers/GroupsRouter.tsx
```diff
@@ -58,22 +59,22 @@ export class GroupsRouter extends React.PureComponent<GroupsRouterProps> {
         }
 
         const groupProps: Partial<GroupsElementProps> = {
-            localGroups: this.props.groups
+            localGroups: this.props.localGroups
         };
 
-        if (groupSplitExistsAtNode(this.props.nodeToEdit)) {
+        if (hasGroupsRouter(this.props.nodeToEdit)) {
```
ycleptkellan: changed this because i realize `split` isn't used much; seems like we're favoring `router` throughout the repo

src/component/routers/GroupsRouter.tsx
```diff
@@ -87,12 +88,12 @@ export class GroupsRouter extends React.PureComponent<GroupsRouterProps> {
...
 
 const mapStateToProps = ({
-    flowContext: { groups },
+    flowContext: { groups: localGroups },
     flowEditor: { editorUI: { translating } },
     nodeEditor: { nodeToEdit }
 }: AppState) => ({
     translating,
-    groups,
+    localGroups,
```
ycleptkellan: changing this from `groups` to `localGroups` to be more specific about the purpose of the prop. perhaps we should do the same in the store for locals?

src/external/index.test.ts
```diff
@@ -1,43 +1,23 @@
...
-afterAll(() => nock.cleanAll());
-
-describe('Providers: external', () => {
+describe('external', () => {
     describe('getFlow', () => {
-        it('should get a flow', () => {
-            getFlow(flowsEndpoint, flowUUID, false).then(response =>
-                expect(response).toEqual(getFlowResp)
+        it('should get a flow', async () => {
+            expect(await getFlow(flowsEndpoint, colorsFlowResp.results[0].uuid, false)).toEqual(
```
ycleptkellan: getting comfortable w/ `async` - `await` syntax; think it can be more readable in some cases

src/external/index.ts
```diff
@@ -9,28 +9,7 @@ export interface FlowDetails {
     dependencies: FlowDefinition[];
 }
 
-export type GetActivity = (
```
ycleptkellan: we weren't using these types

src/store/actionTypes.ts
```diff
@@ -121,145 +127,118 @@ interface UpdateShowResultNameActionPayload {
...
-export interface UpdateLanguageAction {
-    type: Constants.UPDATE_LANGUAGE;
-    payload: LanguagePayload;
-}
-
-export interface UpdateFetchingFlowAction {
-    type: Constants.UPDATE_FETCHING_FLOW;
-    payload: FetchingFlowPayload;
-}
+export type UpdateTranslatingAction = DuxAction<Constants.UPDATE_TRANSLATING, TranslatingPayload>;
```
ycleptkellan: seems like a great use case for generics

src/store/flowContext.ts
```diff
@@ -98,66 +103,116 @@ export const updateContactFields = (
...
     type: Constants.UPDATE_COMPONENTS,
     payload: {
         components
     }
 });
 
-// Reducer
-export default (state: FlowContext = initialState, action: ActionTypes) => {
+// Reducers
+export const definition = (
```
ycleptkellan: breaking these into individual reducers, which feels better to me from an OCD standpoint but i also observed a coverage increase compared to the single-reducer approach 🤷‍♂️

src/component/actions/ChangeGroups/ChangeGroups.test.tsx
```diff
@@ -1,44 +1,41 @@
...
-                shallow(<ChangeGroupsComp {...{ ...addToGroupsAction, groups }} />),
-                'content'
-            );
+const COMPONENT_TO_TEST = ChangeGroupsComp.name;
 
-            expect(Content.children().length).toBe(4);
-            expect(Content.childAt(3).text()).toBe('...');
+describe(`${COMPONENT_TO_TEST}`, () => {
+    describe('render', () => {
+        it(`should render ${COMPONENT_TO_TEST} with group name`, () => {
```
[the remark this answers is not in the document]
ycleptkellan: Yeah, maybe this is lazy/unnecessary?

__mocks__/axios.ts
ycleptkellan: if the `__mocks__` directory lives at the same level as `node_modules`, it'll automatically replace imported node modules w/ mock files of the same name. we can also configure jest to look elsewhere if we wanted to move `__mocks__` 👍

src/component/SelectSearch.test.tsx
ycleptkellan: eh, if we all get the picture i figure the comment above `yieldToPromises` is informative enough. wasn't sure if this was a new or confusing pattern but i get the sense it isn't

src/testUtils/index.tsx
ycleptkellan: i saw this elsewhere but the use of the term 'flush' was a bit confusing for me; saw this as more of a deferral. happy to use if it's a common thing 👍

src/testUtils/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export interface QueryString {
+    [key: string]: string;
+}
+
+/**
+ * Compose setup method for component tests
+ */
+export const createSetup = <P extends {}, C extends {} = { [contextType: string]: any }>(
```
ycleptkellan: i noticed the use of a 'setup' method [here](https://redux.js.org/recipes/writing-tests) and [here](https://medium.com/selleo/testing-react-components-best-practices-2f77ac302d12), but figured making this a higher-order function is more dry (alternative being definition a 'setup' method in each test file). we also need to be able to pass/read from context.

src/component/SelectSearch.test.tsx
ericnewcomer: That there’s a yield when it’s not obvious there should be a request is the comment worthy bit.

ycleptkellan: ah gotcha

PR #214, opened 2018-03-21
fix localization

ycleptkellan:


ycleptkellan:
@ericnewcomer noticed this component wasn't updating when a new language is selected since we're overwriting the `shouldComponentUpdate` lifecycle hook `connect` provides to the component it outputs. Using `isEqual` is a bit heavy-handed so feel free to add more granular checks, just thought we might want to encourage updates on any prop change for `connect`ed components for the time being and optimize later. LMK what you think.

PR #222, opened 2018-03-29
Move to map based state for nodes

ericnewcomer:
Kind of a big shift here.

* Remove the `components` from state
* Empty out the `nodes[]` on definition in state
* Add a node map called `nodes` at the root of state, which maps object  of type `RenderNode` which is our editor rendition of nodes that contains `ui`, `node`, and `incomingConnections`
* source ids have node id's embedded in them `node_uuid:exit_uuid` which is a necessity for working with inbound jsplumb connections and not maintaining our internal mapping.

Trickiest bit and likely biggest surface area for bugs is managing the `incomingConnections` to ensure that we do the right thing there. This complexity should be encapsulated in `thunks.js` and will have thorough tests in time (I've made a start here).


src/store/flowEditor.test.ts
```diff
@@ -400,9 +400,9 @@ describe('flowEditor reducers', () => {
             const { definition } = colorsFlow.results[0];
             const fromNode = definition.nodes[0];
             const fromNodeUI = getNodeUI(fromNode.uuid, definition);
-            const ghostNode = getGhostNode(fromNode, fromNodeUI, definition);
-            const action = updateGhostNode(ghostNode);
-            expect(reduce(action)).toEqual(ghostNode);
+            // const ghostNode = getGhostNode(fromNode, fromNodeUI, nodes);
+            // const action = updateGhostNode(ghostNode);
+            // expect(reduce(action)).toEqual(ghostNode);
```
ericnewcomer: same here

src/store/thunks.test.ts
```diff
@@ -0,0 +1,82 @@
+jest.unmock('redux-mock-store');
+jest.unmock('immutability-helper');
+
+// import {createMockStore} from 'redux-mock-store';
+const createMockStore = require('redux-mock-store');
```
ycleptkellan: JC why the move to CommonJS syntax?

ycleptkellan: NVM i see why

src/component/NodeEditor/NodeEditor.tsx
```diff
@@ -1208,9 +1208,9 @@ const mapStateToProps = ({
     language,
     nodeEditorOpen,
     actionToEdit,
-    localizations,
     definition,
```
ycleptkellan: thinking we could use `localizations` state for `cleanUpLocalizations` and the update router methods, and lose `definition` state as a dependency; what do you think? my bias being to not read from `definition` state if we can help it

PR #225, opened 2018-03-29
Component tests

ycleptkellan:
- adds tests for a few components 
- moves us back to the current context API to make tests easier to write/follow
- updates jest 
- shallow renders if possible, mounts only if necessary (e.g. asserting our widget-binding methods are called)
- tests wrapped components instead of those returned by `connect`, mocks out action creators


ericnewcomer:
This approach seems sane to me

ycleptkellan:
@ericnewcomer i can't figure out the push failure, any ideas as to what might be causing that? this is ready otherwise

ericnewcomer:
Would really like to see us depend on a more concise flow just for tests, but this is a great start.

PR #228, opened 2018-03-29
Minified jsplumb

ycleptkellan:


ericnewcomer:
During dev, I don't see the value in using the minified version, doesn't this get minified when we bundle up for production anyways? If an error happens inside jsPlumb code, I don't want to have a giant blob of obfuscated json that doesn't help me.

ycleptkellan:
ah didn't realize it was obscuring errors, closing

Issue #224, opened 2018-03-29
Reimplement reflow


ericnewcomer:
Now that we aren't sorting nodes on mutation, need to come up with a plan for reflow

PR #236, opened 2018-04-03
Reflow

ericnewcomer:
* Store bounds (left, right, top, bottom) on position instead of separate dimension object on node.ui
* Do one reflow collision at a time, recurse as necessary
* Rework thunk tests to use shared ad-hoc nodes object instead of sample flow
* Only reflow on node move or bounds change

Fixes: #224


ycleptkellan:
ah, gotcha. FWIW i think both could actually benefit from a slower transition at some point; they're cool effects and i think it would punctuate both nicely.

PR #237, opened 2018-04-03
Fix gridding and parameterize it

ericnewcomer:
I like this better since it makes the snapping really tight in all cases, thoughts?


ericnewcomer:
I like this better since it makes the snapping really tight in all cases, thoughts?

ycleptkellan:
I could see a smaller grid being helpful for that reason and I notice it's also creating a more pronounced contrast between the canvas and the nodes (now that they aren't given a shadow while resting). That being said, this size feels a tad too small. Accentuating the snapping (perhaps by slowing the transition a little) could add character, make dragging and dropping nodes more pleasant.

ycleptkellan:
Ah, gotcha. Not a big deal, but i'll give that a try.

nicpottier:
I think I agree with Kellan that maybe this is a bit small, at least I don't seem to be aligning nodes very easily on the first try. (which is the whole point of snapping ya?)

ericnewcomer:
I think I agree a bit bigger would be better. Just couldn’t easily make it work quickly without it bein divisible by 10 :)

ericnewcomer:
Okay, sticking with grid size of 20 (_same as its always been, same as it ever was_). Would still like to merge this since I've colocated all the gridding variables and fixed some gridding math that was definitely off but only worked coincidentally.

ericnewcomer:
Ha, me either, but really wanted it. Very well might be an anti-pattern tho.

PR #252, opened 2018-04-06
Fix 250

### LINKED ISSUE #250: SetContactAttrib should prune whitespace
Both when creating a new field and checking for reserved names.

ycleptkellan:
fixes #250, brings `AttribElement` up to 100% line, branch coverage


src/component/SelectSearch.tsx
```diff
@@ -16,10 +16,11 @@ export interface SelectSearchProps {
     closeOnSelect?: boolean;
     initial?: SearchResult[];
     localSearchOptions?: SearchResult[];
-    _className?: string;
+    __className?: string;
     createPrompt?: string;
     onChange?: (selections: SearchResult | SearchResult[]) => void;
     isValidNewOption?: (option: { label: string }) => boolean;
+    isOptionUnique?: Function;
```
ericnewcomer: Isn't this function a consistent signature?

src/component/form/AttribElement.tsx
```diff
@@ -1,37 +1,56 @@
...
 }
 
 export const PLACEHOLDER = 'Enter the name of an existing attribute or create a new one';
 export const NOT_FOUND = 'Invalid attribute name';
+export const VALID_FIELD = /^[a-z0-9-][a-z0-9- ]*$/;
+export const CREATE_PROMPT = 'New attribute: ';
+
+export const attribExists = (newOptName: string, options: SearchResult[]) => {
+    const loweredNTrimmed = newOptName.toLowerCase().trim();
+    if (options.length) {
```
ericnewcomer: for things like this, what about instead doing 
```javascript
options.find((searchResult: SearchResult)=> ...
```
...would remove a branch in coverage too

src/component/form/AttribElement.tsx
```diff
@@ -1,37 +1,56 @@
...
+                return true;
+            }
+        }
+    }
+    return false;
+};
+
+export const fieldNameValid = (name: string = '') => {
+    const lowered = name.toLowerCase();
+    return lowered.length > 0 && lowered.length <= 36 && VALID_FIELD.test(lowered);
```
ericnewcomer: We should have global config somewhere, been putting them in utils (which is where i export the ones from scss land). Would be nice to not see a 36 buried in our code.

src/component/form/FormElement.tsx
```diff
@@ -19,19 +20,18 @@ export interface FormElementProps {
 
 export default class FormElement extends React.PureComponent<FormElementProps> {
     private getName(): JSX.Element {
-        if (this.props.showLabel && this.props.name) {
-            return <div className={styles.label}>{this.props.name}</div>;
-        }
-
-        return null;
+        return renderIf(
```
ericnewcomer: neat!

src/component/SelectSearch.tsx
ycleptkellan: Yes that was lazy of me 😅. While addressing this I realized we could be leaning on `react-select` types a lot more that we do. What are your thoughts on https://github.com/nyaruka/floweditor/pull/252/commits/7d526f166f06300642a6c31ec9c6ae147d6ab817?

src/component/form/AttribElement.tsx
ycleptkellan: much cleaner, addressed here https://github.com/nyaruka/floweditor/pull/252/commits/31b503b1cca54427639dcf70ad6cda885b3b0775

src/component/form/AttribElement.tsx
ycleptkellan: Cleaned this up a bit more here https://github.com/nyaruka/floweditor/pull/252/commits/e1a3e474987cbe9a879d915fff4f9ef81c478eb7. 

I'm fine with them living in utils for now, but certainly not opposed to moving them to another location if you know where you'd like them to be. Not sure what to call them/where in the repo would make the most sense outside of utils.

PR #257, opened 2018-04-10
Embed Tweaks

ericnewcomer:
These are just some layout tweaks to position things more realistically as if they lived inside some container to make sure that we are doing right by jsplumb and any absolute positioning within the editor.

Also, anchored flow editor coordinate space at `0, 0` instead of `GRID_SIZE, GRID_SIZE`


ericnewcomer:
I don’t care if it’s chartreuse. The purpose is to not be anchored at the top of the page to make sure we aren’t making positional assumptions.

ericnewcomer:
Also, why does it matter if it's a similar color to the TextIt brand? I don't understand the concern.

ycleptkellan:
Totally understand the goal here, just thought it would be nice to change things up. NBD

PR #258, opened 2018-04-11
Fix selector

ericnewcomer:
Honor the parent offset for drag selection


ericnewcomer:
We really need to get some baseline tests on these components

ycleptkellan:
I'll submit a PR for `Flow` coverage 👍

Issue #93, opened 2017-12-20
Add sticky notes


ericnewcomer:
We need to implement the old version of sticky notes on double click.

One thing that is sorely lacking in the old editor is a way to create brand new nodes which can currently only be done by dragging from an existing node which isn't always possible or prudent on a live flow. One thought is that instead of immediately creating the sticky note on double click, we could present a context menu in place that shows options to either create a new note or a new node.

ycleptkellan:
Thats a cool idea. Wondering what sorts of ideas you have for the context menu component. I'm picturing a popover at click location that presents two choices, node (opens node editor) or note (renders a note component).

ericnewcomer:
Discussed this a bit further with @nicpottier and wanted to capture that discussion here. The result of that discussion is that the context menu is less important with our ability to add and splice nodes into the flow present in the new editor. Adding a context menu maybe very well add unnecessary complexity for the user, especially since we are way nicer about adding actions or splits to existing nodes without forcing them to drag.

ycleptkellan:
Want to make sure I'm following, what's the plan for making adding actions and splits nicer? Does this mean we aren't going to provide a way to add a new node without dragging from an existing node? 

Also, what conclusion did you come to about discoverability? Seems to me like one of the least obvious features in the editor, but then again I do see flows with sticky notes in the wild...

ycleptkellan:
Ah, I understand now thank you. That might be a good question to pose to power users, would it make sense to set up a feedback-gathering process for things like this?

ericnewcomer:
Post on http://feedback.textit.in ?

ycleptkellan:
Worth a try, though I'm hoping we'll reconsider UserVoice at some point. I'll make a note to look at Intercom's offerings/suggestions as well JIC.

PR #262, opened 2018-04-12
Stickies

ericnewcomer:
## WIP: Sticky Notes
- [x] Render existing sticky notes
- [x] Update definition on position change
- [x] Update definition on text change with debounce
- [x] Create new sticky notes on double click
- [x] Remove sticky notes
- [x] Color options
- [x] Store unit tests
- [x] Component unit tests

Okay, this one is ready to go. Fixes: #93


src/component/Flow.test.ts
```diff
@@ -22,12 +22,18 @@ import Plumber from '../services/Plumber';
...
 import { getCollisions, getGhostNode, getRenderNodeMap } from '../store/helpers';
 import { createSetup, createSpy, getSpecWrapper } from '../testUtils';
-import { getBaseLanguage } from '../utils';
+import { getBaseLanguage, dump } from '../utils';
 
 jest.mock('../services/ActivityManager');
 jest.mock('../services/Plumber');
 jest.useFakeTimers();
 
+jest.mock('uuid', () => {
```
ycleptkellan: this looks like learning opportunity, why might we want to hijack `v4`/`generateUUID`?

ycleptkellan: in part so that we can generate snaps?

src/component/Sticky/Sticky.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { getBaseLanguage, dump } from '../../utils';
+import { ConnectionEvent, createStore, initialState } from '../../store';
+import { object } from 'prop-types';
+import { endpointsPT, languagesPT } from '../../config/ConfigProvider';
+import * as styles from './Sticky.scss';
+import * as config from '../../../assets/config';
+
+jest.useFakeTimers();
+
+// TODO: this is an awful lot of boilerplate for testing a simple component
```
ycleptkellan: this doesn't feel _too_ bad to me (makes sense to be explicit about prop requirements, context subscriptions). wonder if having `context` and `childContextTypes` initialized by default in `createSetup` wouldn't help? that's the most laborious part for me, anyways, and there's no harm in passing all of `context` to components that don't read from it.

src/component/Sticky/Sticky.tsx
```diff
@@ -0,0 +1,250 @@
...
+        this.props.onResetDragSelection();
+    }
+
+    private onDrag(event: DragEvent): void {
+        // noop
+    }
+
+    private onDragStop(event: DragEvent): void {
+        // snap us to the same grid
+        const { left, top } = snapToGrid(event.finalPos[0], event.finalPos[1]);
```
ycleptkellan: JW, since stickies can be placed anywhere on the canvas...aren't really restricted in any way...should we be snapping them to the grid? is that something you had considered? OCD ALERT: i played with dragging a sticky over a node (something i'd probably do if i were on a team sharing an account), and was disappointed that it was snapping to the canvas.

src/component/Sticky/Sticky.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+        it('should update the body', () => {
+            updateText('body', 'My new body');
+        });
+
+        it('should debounce multiple updates', () => {
+            updateText('title', 'Update one');
+            updateText('title', 'Update two');
+
+            // TODO: make clearTimeout mock work
+            // expect(clearTimeout).toHaveBeenCalled();
```
ycleptkellan: i think you'd need to call these sequentially, e.g.

```
const firstUpdate = 'Update one';
const secondUpdate = 'Update two';
const selector = 'textarea.title';
const { wrapper, instance } = setup({}, false);

wrapper.find(selector).prop('onChange')({
	currentTarget: { value: firstUpdate }
});

expect(wrapper.find(selector).text()).toBe(firstUpdate);

wrapper.find(selector).prop('onChange')({
   currentTarget: { value: secondUpdate }
});

expect(wrapper.find(selector).text()).toBe(secondUpdate);

expect(clearTimeout).toHaveBeenCalledTimes(1);
```

works for me

src/component/Flow.test.ts
ericnewcomer: this is just for deterministic tests, and ya, snaps especially

src/component/Sticky/Sticky.tsx
ericnewcomer: If you check the todos, future plans here are to have notes play in the same space as nodes. ie, you can insert them in between nodes and they will reflow, can drag select and move together with nodes, etc

ycleptkellan: But does that require them to snap? If so, why?

src/component/Sticky/Sticky.test.tsx
ericnewcomer: ha, it's because my wrapper setup was inside my helper function, thereby resetting it

src/component/Flow.test.ts
ycleptkellan: should we add this to `__mocks__`? or do you see this as something we'd want to do in select cases?

src/component/Sticky/Sticky.test.tsx
ericnewcomer: the drier the better as far as im concerned. need to keep the tests as concise and just meat as possible.

src/component/Flow.test.ts
ericnewcomer: I imagine tests will want to reset the counter at various points. Is there a way to get it to reset itself without putting it in `beforeEach`?

src/component/Sticky/Sticky.tsx
ericnewcomer: Not sure why they would play differently than nodes in that case. Especially if you are inserting them in between nodes and reflowing.

ycleptkellan: Yeah I guess that would be a req. How committed are we to putting those on the same plane as nodes? Would that be resolving an issue peeps are running into in the v1 editor?

ericnewcomer: Pretty committed, just feels more natural being able to insert notes and the flow making room as necessary just as it does with the rest of the flow. That said, I don't see it as an essential thing out the gate.

src/component/Flow.test.ts
ycleptkellan: we can reset it between each test using the [clearMocks](https://facebook.github.io/jest/docs/en/configuration.html#clearmocks-boolean) setting. can verify that works w/ our axios mock in `__mocks__`.

src/component/Sticky/Sticky.test.tsx
ycleptkellan: ok probably worth opening a PR for that. i can make that happen when this is merged, unless you want to tack that on.

src/component/Sticky/Sticky.tsx
ycleptkellan: Yeah, fair. Guess it does make sense to really lean into the reflow paradigm here.

ericnewcomer:
Been playing a bit with universal mocking of uuid generation, but resetMocks in jest config doesn't seem to do what we want there. We should probably merge this and look at reducing boilerplate setup and universal mocking of uuid as separate PRs.

ycleptkellan:
Aye

PR #267, opened 2018-04-13
Boring Pull Request

ericnewcomer:
* Add new flow to test with `boring.json`
* Remove all test flows in `store/__test__.ts` and refactor them to be boring (which itself was a laboriously boring task)
* Add copy of boring to the preview options just for SAGs so we can visualize it if necessary when writing tests.

Boring is simple four node flow with predetermined values for uuid that are predictable (`node0`, `node3_action0`, `node2_exit0`, etc). This makes tests simpler and cleaner.


lambda-src/flows.js
```diff
@@ -9,24 +8,23 @@ const baseOpts = {
...
-    return flowResp ? flowResp : false;
+    }
+    return false;
 };
 
 const notFoundHandler = cb => cb(null, getOpts({ statusCode: 404, body: signBunny('not found') }));
 const flowsHandler = ({ queryStringParameters: query } = {}, cb) => {
     if (Object.keys(query).length > 0) {
-        if (query.uuid && isValidUUID(query.uuid)) {
+        if (query.uuid) {
```
ycleptkellan: `isValidUUID` was tripping us up

PR #269, opened 2018-04-16
Remove more references to non-test assets

ericnewcomer:
Removes references to colors and customer service flows for all of the tests in the store module (and any references to non-test assets). Also refactors a few other tests as an example for use to more toward (`Action.test.ts`, `GroupRouter.test.tsx`, etc).

### DIFF
--- a/src/component/actions/Action/Action.test.ts
+++ b/src/component/actions/Action/Action.test.ts
@@ -9,16 +9,48 @@ import {
     actionInteractiveDivSpecId,
     actionBodySpecId
 } from './Action';
+import { FlowNode, SendMsg, StartFlow, SwitchRouter } from '../../../flowTypes';
 
-const config = require('../../../../assets/config');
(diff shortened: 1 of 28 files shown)


src/component/actions/Action/Action.test.ts
```diff
@@ -9,16 +9,48 @@ import {
...
+    exits: [],
+    router: {
+        type: 'switch',
+        operand: '@child'
+    } as SwitchRouter,
+    wait: {
+        type: 'flow',
+        flow_uuid: 'flow_to_start'
+    }
+};
```
ericnewcomer: We should be following this pattern for action (and other unit level tests). Instead of pulling out bits and pieces from some complex flow definition, lets just define what we are going to test with inline and keep it simple. This reduces dependency entanglements and is much clearer at the small expense of some verbosity.

ycleptkellan: agreed, was just working on some helpers for this sort of thing

ericnewcomer:
woot

PR #270, opened 2018-04-17
Jesticulate

ycleptkellan:
- adds/modifies existing testing utilities w/ the aim of reducing testing boilerplate, e.g. `genSendMsgAction`, `genFlowNode`, etc.
- types 'Types', e.g. `send_msg`...not sure what to call those hehe.
- types operators.
- adds convenience methods for `immutability-helper` queries, e.g. `{ $set: { text: 'Hi there!' } }` becomes `set({ text: 'Hi there!' })`...mostly a personal bias there, happy to remove.
- makes jest aware of `matchers.ts` so we don't need to import it.


src/testUtils/index.tsx
```diff
@@ -15,31 +47,79 @@ export interface QueryString {
...
+            ...initialState.flowEditor,
+            editorUI: {
+                ...initialState.flowEditor.editorUI,
+                language: getBaseLanguage((config as FlowEditorConfig).languages)
+            }
+        }
+    }
+};
+
+export const configProviderContext: ConfigProviderContext = {
```
ycleptkellan: exporting an object here that contains only the configuration properties the app cares about

src/testUtils/index.tsx
```diff
@@ -15,31 +47,79 @@ export interface QueryString {
...
+};
+
+export const setMock = (implementation?: (...args: any[]) => any): Query<jest.Mock> =>
+    set(jest.fn(implementation));
+
 /**
  * Compose setup method for component tests
  */
-export const createSetup = <P extends {}, C extends ConfigProviderContext = ConfigProviderContext>(
+export const composeSetup = <P extends {}>(
```
ycleptkellan: finally updating this bad boy. modified to allow us to update `store` state, utilize `immutability-helper`. 'compose' because it's a higher-order function.

src/testUtils/index.tsx
```diff
@@ -79,20 +160,309 @@ export const getSpecWrapper = (
...
-};
+): any => componentWrapper.find(`[${attributeName}="${specName}"]`);
 
-const DEBUG = false;
+export const composeDuxState = (
+    query: Query<AppState | Partial<AppState>> = {},
+    duxState = baseState
+) => mutate(duxState, query);
 
-export const timeStart = (name: string) => {
```
ycleptkellan: moved these to `/utils` since we were importing them into the app/not using them for tests

src/testUtils/index.tsx
ycleptkellan: forgot to mention it also passes context, store by default so we don't have to initialize them in each test

src/testUtils/index.tsx
```diff
@@ -79,20 +160,309 @@ export const getSpecWrapper = (
...
-    if (DEBUG) {
-        console.time(name);
-    }
-};
+export const composeComponentTestUtils = <P extends {}>(
```
ycleptkellan: adding a method to tailor `setup` and `spyOn` to the component being tested

src/testUtils/index.tsx
```diff
@@ -79,20 +160,309 @@ export const getSpecWrapper = (
...
+    // tslint:disable-next-line:variable-name
+    all_urns?: boolean;
+} = {}): SendMsg => ({
+    type: Types.send_msg,
+    uuid,
+    text,
+    all_urns
+});
+
+export const genSendEmailAction = ({
```
ycleptkellan: adding initialization methods for each component we're testing. i like that we can use default parameters to return 'default versions' of each item as opposed to keeping them nearby in a `__test__` file, but happy to remove if you think this is overkill 🤙

src/component/index.tsx
```diff
@@ -73,18 +72,18 @@ export class FlowEditor extends React.Component<FlowEditorStoreProps> {
...
                 id={editorContainerSpecId}
-                className={translatingClass}
+                className={this.props.translating ? styles.translating : undefined}
                 data-spec={editorContainerSpecId}
             >
                 <div className={styles.editor} data-spec={editorSpecId}>
                     <ConnectedLanguageSelector />
                     <ConnectedFlowList />
-                    {renderFlow && <ConnectedFlow />}
+                    {renderIf(
```
ycleptkellan: trying to make `renderIf` a thing...

src/component/Flow.tsx
```diff
@@ -73,7 +73,7 @@ export const dragSelectSpecId = 'drag-select';
 
 export const getGhostUI = (ghostNode: FlowNode = {} as any) => ({
     position: GHOST_POSITION_INITIAL,
-    ...(ghostNode.router ? { type: 'wait_for_response' } : {})
+    ...(ghostNode.router ? { type: Types.wait_for_response } : {})
```
ericnewcomer: Agree on using constants for these, but they sort of feel like they should be with flowTypes since they are describing the internals of the flow definition, but I see you have operators too. Those have a similar vibe to me, hmmm.. we can chat about that later I spose.

src/component/Flow.tsx
```diff
@@ -257,18 +257,13 @@ export class Flow extends React.Component<FlowStoreProps, {}> {
     }
 
     private getStickies(): JSX.Element[] {
-        let stickyMap = this.props.definition._ui.stickies;
-        if (!stickyMap) {
-            stickyMap = {};
-        }
-
+        const stickyMap = this.props.definition._ui.stickies || {};
```
ericnewcomer: 😎always forget this pattern

src/component/SelectSearch.test.ts
```diff
@@ -1,38 +1,41 @@
...
-
-describe(`${COMPONENT_TO_TEST}`, () => {
     describe('render', () => {
         it('should render self, children with required props', async () => {
-            const selectRefSpy = jest.spyOn(SelectSearch.prototype, 'selectRef');
-            const loadOptionsSpy = jest.spyOn(SelectSearch.prototype, 'loadOptions');
-            const searchSpy = jest.spyOn(SelectSearch.prototype, 'search');
+            const selectRefSpy = spyOn('selectRef');
+            const loadOptionsSpy = spyOn('loadOptions');
+            const searchSpy = spyOn('search');
```
ericnewcomer: automagic component spyOn is neat 👍

src/component/actions/Action/Action.test.ts
```diff
@@ -1,67 +1,44 @@
...
-        operand: '@child'
-    } as SwitchRouter,
-    wait: {
-        type: 'flow',
-        flow_uuid: 'flow_to_start'
+const sendMsgAction = genSendMsgAction();
+const sendMsgAction1 = genSendMsgAction({ uuid: 'send_msg-1', text: 'Yo!' });
+const sendMsgNode = genFlowNode({ actions: [sendMsgAction], exits: [genExit()] });
+const startFlowAction = genStartFlowAction();
+const startFlowNode = genStartFlowNode(startFlowAction);
```
ericnewcomer: This makes a lot of sense and is pretty DRY. Especially if we end up changing the flow format, this will be much easier to maintain. Can we call them `createBlah` instead of `gen`? Maybe they alos deserve their own file in the test utils.

src/store/__test__.ts
```diff
@@ -0,0 +1,111 @@
+import { Operators } from '../config/operatorConfigs';
```
ericnewcomer: this file was removed previously and was replaced by boring. Maybe a bad merge brought it back from the dead.

src/store/flowContext.test.ts
```diff
@@ -19,32 +19,37 @@ import reducer, {
...
     RenderNodeMap
 } from './flowContext';
+import { configProviderContext } from '../testUtils';
+import { FlowDefinition } from '../flowTypes';
+import { Types } from '../config/typeConfigs';
 
-const config = require('../../assets/config');
-const definition = require('../../__test__/flows/boring.json');
+const colorsFlow = require('../../__test__/flows/colors.json') as FlowDefinition;
+const customerServiceFlow = require('../../__test__/flows/customer_service.json') as FlowDefinition;
```
ericnewcomer: Why bring these back instead of `boring`? Bad merge?

src/store/flowEditor.test.ts
```diff
@@ -24,17 +28,12 @@ import {
...
+    updateTranslating
 } from './flowEditor';
 import { getGhostNode } from './helpers';
-import { FlowNode } from '../flowTypes';
-import { RenderNode } from './flowContext';
 
-const config = require('../../assets/config');
-const flowsResp = require('../../__test__/assets/flows.json');
-const definition = require('../../__test__/flows/boring.json');
+const flowsResp = require('../../assets/flows.json');
```
ericnewcomer: `__test__/assets/...`

src/store/flowEditor.test.ts
```diff
@@ -24,17 +28,12 @@ import {
...
+const colorsFlow = require('../../__test__/flows/colors.json');
```
ericnewcomer: what do you have against boring?!

src/store/mutators.ts
```diff
@@ -2,13 +2,14 @@ const mutate = require('immutability-helper');
...
-import { dump, snapToGrid } from '../utils';
+import { dump, snapToGrid, set, merge, unset, splice } from '../utils';
 import { getUniqueDestinations, getNode, getExitIndex, getActionIndex } from './helpers';
 import { LocalizationUpdates } from '.';
+import { push } from '../utils/index';
 
 export const uniquifyNode = (newNode: FlowNode): FlowNode => {
     // Give our node a unique uuid
-    return mutate(newNode, { $merge: { uuid: generateUUID() } });
+    return mutate(newNode, merge({ uuid: generateUUID() }));
```
ericnewcomer: functions named `set`, `merge`, etc feel really overloaded, but willing to give it a go.

src/testUtils/index.tsx
```diff
@@ -15,31 +47,79 @@ export interface QueryString {
...
+    ...{
+        flowEditor: {
+            ...initialState.flowEditor,
+            editorUI: {
+                ...initialState.flowEditor.editorUI,
+                language: getBaseLanguage((config as FlowEditorConfig).languages)
+            }
+        }
+    }
+};
```
ericnewcomer: i know this existed before this PR but man, I just needed to say this is spread on some hardcore drugs

src/testUtils/index.tsx
```diff
@@ -79,20 +160,309 @@ export const getSpecWrapper = (
...
+        id: uuid
+    }));
+
+export const getGroups = (sliceAt: number, groups: Group[] = groupsResults) =>
+    groups
+        .map(({ name, uuid }) => ({
+            name,
+            id: uuid
+        }))
+        .slice(sliceAt);
```
ericnewcomer: Really like this approach, see comments above

jest.config.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
     collectCoverageFrom: ['src/**/*.{ts,tsx}'],
     coverageReporters: ['lcov', 'json'],
     coveragePathIgnorePatterns: ['/*.d.ts$', 'lambda-src', 'webpack', 'testUtils', '__test__.ts'],
-    setupFiles: ['<rootDir>/setup.js']
+    setupFiles: ['<rootDir>/src/testUtils/setup.ts'],
+    setupTestFrameworkScriptFile: '<rootDir>/src/testUtils/matchers.ts'
```
ericnewcomer: 👍

src/component/Flow.tsx
ycleptkellan: Open to anything here, just wanted to ensure we type these before too long

src/component/actions/Action/Action.test.ts
ycleptkellan: Totally, moved these to `assetCreators` and gave them the 'create' prefix here https://github.com/nyaruka/floweditor/pull/270/commits/d86214d4fb59dfa9385c9917c57c03ec0d2f453f

src/store/mutators.ts
ericnewcomer: Ha. Overloaded is somewhat overloaded itself ;) Generally it is used to refer to commonly named things that can confuse meaning, so for example, would also say "Action" is pretty overloaded in this codebase :)

src/testUtils/index.tsx
```diff
@@ -15,31 +19,71 @@ export interface QueryString {
...
+export const baseState: AppState = mutate(initialState, {
+    flowContext: merge({
+        definition: require('../../__test__/flows/colors.json') as FlowDefinition
+    }),
+    flowEditor: {
+        editorUI: merge({
+            language: getBaseLanguage((config as FlowEditorConfig).languages)
+        })
+    }
+});
```
ericnewcomer: 💯

src/store/mutators.ts
ycleptkellan: ahhh, gotcha thanks. perhaps we should change these to `setImmutably`, `mergeImmutably`, etc?

PR #274, opened 2018-04-19
Declare custom matcher types

ycleptkellan:
...so we don't have to worry about importing custom matchers/custom matcher types.


ericnewcomer:
Does this appease the vscode gods somehow? Don't see how it would be any different from it's perspective. Do we need typings too somewhere?

ycleptkellan:
No this should do the trick (does for me at least). `setupFiles` is where we'd make any global declarations (required before the test env is initialized; jest object isn't a thing yet) while `setupTestFrameworkScriptFile` is more appropriate for things like custom matcher declarations (called before each individual test runs; global jest object exists at that point). That's my understanding anyways.

PR #281, opened 2018-04-23
AssetService for Groups

ericnewcomer:
So here's a naive asset service just for groups in memory. You'll notice the add group action doesn't take a URL endpoint or local options, instead it's passed an Assets object which will serve as our abstraction with the real asset server and local storage.

- [x] Use AssetService when user adds a group through the group widget
- [x] Add groups to asset service on definition initialization 
- [x] Configure for add to group
- [x] Include faked assets 
- [x] Remove dependency on prop-types 
- [x] Call out to an endpoint based on configuration
- [x] Local results should return promises for consistency with asset server 
- [x] Configure for all places where groups are used (group split, remove)


src/services/AssetService.ts
```diff
@@ -0,0 +1,47 @@
...
+    name: string;
+}
+
+export interface Assets {
+    search(term: string): SearchResult[];
+    add(searchResult: SearchResult): void;
+    addAll(searchResults: SearchResult[]): void;
+}
+
+export class GroupAssets {
```
ycleptkellan: what are your thoughts on logging each change to `GroupAssets`' state while in development? e.g. 

```
prevState: {
	groups: []
},
nextState: {
	groups: [{}, {}, {}],
}
```

i don't like not being able to inspect a singleton's state, but perhaps that's not a shared concern/there's a better approach than logging each change?

ericnewcomer:
Interesting thought...

I think this depends on where things live. We could back the AssetService
in redux, but I think ultimately we want to put it into local storage. I
think if state management/changes/monitoring was the goal, redux is what
you want but doesn’t fit the goal of non ephemeral assets.

By using local storage we would also have the option to not have everything
in memory all the time. I suppose we could create a simple component to
list and edit what’s in our asset server (that’s not part of the editor
component we ship).

ycleptkellan:
Sounds good to me, I don’t imagine these services getting too wild anyways.   
  

  
I did look into ‘redux-persist’ but that might be too heavy-handed for our needs.   
  

  
>

src/component/Node.scss
```diff
@@ -85,7 +85,6 @@
         }
 
         .actions {
-            width: $node_width;
```
ericnewcomer: not related, two demerits. this was breaking the webhook width issue

src/component/__snapshots__/SelectSearch.test.ts.snap
```diff
@@ -1,6 +1,34 @@
...
+  "name": "Groups",
+  "onBlurResetsInput": true,
+  "onChange": [Function],
+  "openOnFocus": true,
+  "options": Array [],
+  "placeholder": "Enter the name of an existing group...",
+  "searchPromptText": "Enter the name of an existing group",
+  "searchable": true,
+  "value": undefined,
+  "valueKey": "id",
```
ericnewcomer: snapshots are a good fit for asserting props

src/component/actions/Action/Action.scss
```diff
@@ -11,11 +11,23 @@
...
+    padding: $padding;
     line-height: 1.4em;
     word-wrap: break-word;
     width: auto;
+
+    &.send_msg, &.set_contact_field, &.set_contact_property, 
+    &.add_contact_groups, &.remove_contact_groups, &.set_run_result, 
+    &.send_email {
+        width: $node_width - $padding * 2;
+    }
```
ericnewcomer: other half of the webhook width fix

webpack/webpack.dev.js
```diff
@@ -40,24 +40,6 @@ const devConfig = {
...
-                pathRewrite: { '^/rapid': '' },
-                changeOrigin: true,
-                secure: false
-            },
-            '/textit': {
-                target: 'https://textit.in/api/v2/',
-                pathRewrite: { '^/textit': '' },
-                changeOrigin: true,
-                secure: false
-            },
```
ericnewcomer: we don't and likely won't be using these proxies anymore

PR #282, opened 2018-04-24
UMD build config for NPM

ycleptkellan:
Adds a script, `build:umd`, that outputs minified JS, CSS files to be published to NPM.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,18 +1,18 @@
 {
   "compilerOptions": {
+    "watch": false,
     "moduleResolution": "node",
     "jsx": "react",
     "noImplicitAny": true,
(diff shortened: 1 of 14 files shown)


tsconfig.json
```diff
@@ -1,18 +1,18 @@
...
-    "outDir": "./dist/",
+    "outDir": "lib",
     "preserveConstEnums": true,
     "removeComments": false,
-    "target": "es6",
-    "module": "es6",
-    "allowJs": true,
-    "watch": true,
+    "target": "es5",
+    "downlevelIteration": true,
```
ycleptkellan: We need to make the `downlevelIteration` setting true if we're using iterators like `Array.prototype.entries()`.

PR #285, opened 2018-04-25
Draggable simulator

ericnewcomer:


src/component/Flow.test.ts
```diff
@@ -20,6 +20,7 @@ import { ConnectionEvent } from '../store';
 import { getFlowComponents, getGhostNode } from '../store/helpers';
 import { composeComponentTestUtils, composeDuxState, getSpecWrapper, setMock } from '../testUtils';
 import { merge, set, setTrue, dump } from '../utils';
+import toJson from 'enzyme-to-json';
```
ycleptkellan: looks like we can lose this guy

src/component/Flow.test.ts
```diff
@@ -209,14 +210,11 @@ describe(Flow.name, () => {
...
-                contactFields: [],
-                groups: [],
-                nodes: {}
-            });
-
+            // if we let jest string this one for us it fails
+            // presumably it's hitting an OOM due to circular references
+            // and reporting it as a RangeError
+            // https://github.com/nodejs/node-v0.x-archive/issues/14170
+            expect(JSON.stringify(wrapper.find('Simulator').props(), null, 1)).toMatchSnapshot();
```
ycleptkellan: 😫

src/services/Plumber.ts
```diff
@@ -107,23 +107,34 @@ export default class Plumber {
 
     public draggable(
         uuid: string,
-        start: Function,
-        drag: Function,
-        stop: Function,
-        beforeDrag: Function
+        start: Function = null,
```
ycleptkellan: just a thought for future development from a jsPlumb n00b: we might consider passing this method an options object so it's clearer on call which event handler a passed function is intended for.

ycleptkellan: of course, typing `plumberDraggable` would also help for those of us using vscode

src/component/Flow.test.ts
ericnewcomer: oops, right.

src/component/Flow.test.ts
ericnewcomer: 🤢 ... but oh well

src/services/Plumber.ts
ericnewcomer: Feel like the plumber instance could alternatively just be context-ized

ycleptkellan: Ya perhaps that's a longer-term goal, migrating our services to the new context API?

ericnewcomer: Ya, perhaps :)

ericnewcomer:
Double approval!! Thanks @ycleptkellan!

ycleptkellan:
Rough internet sitch at the moment 😅

ericnewcomer:
Don't be embarrassed by your approval exuberance. Your secret is safe with me.

PR #286, opened 2018-04-25
AssetService for ContactFields Tweaks

ycleptkellan:


preview/src/index.html
```diff
@@ -29,7 +29,7 @@
 <body>
     <div id="nav-bar"></div>
     <div id="flow-editor"></div>
-    <script src="floweditor.js"></script>
+    <script src="flow-editor.js"></script>
```
ycleptkellan: idk if you were running into this problem, but i wasn't able to run the app in development because the filename we were outputting was `@nyaruka/flow-editor.js`...seems like an oversight on my part...should have been addressed by https://github.com/nyaruka/floweditor/pull/283

src/store/helpers.ts
```diff
@@ -242,7 +242,7 @@ export interface FlowComponents {
 }
 
 export const isGroupAction = (actionType: string) => {
-    return actionType === 'add_contact_groups' || actionType === 'remove_contact_groups';
+    return actionType === Types.add_contact_groups || actionType === Types.remove_contact_groups;
```
ycleptkellan: left-over from https://github.com/nyaruka/floweditor/commit/e93beab3092dc599e4fe4ade9dd7c89df8f3bac2

webpack/utils.js
```diff
@@ -0,0 +1 @@
+exports.pkgName = () => require('../package.json').name.replace('@nyaruka/', '');
```
ycleptkellan: utility to strip our scope from the pkg name

PR #287, opened 2018-04-26
AssetService for flows

ericnewcomer:
This one straddles redux a bit. I didn't want to remove the bits that components were relying on in redux until we add redux support to the AssetService all at once.

### DIFF
--- a/src/services/AssetService.ts
+++ b/src/services/AssetService.ts
@@ -1,3 +1,4 @@
+// tslint:disable:max-classes-per-file
 import { SearchResult } from '../store';
 import { Endpoints, FlowEditorConfig, ContactProperties, AttributeType } from '../flowTypes';
 import axios, { AxiosResponse } from 'axios';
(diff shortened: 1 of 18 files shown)


src/services/AssetService.ts
```diff
@@ -6,17 +7,43 @@ import { FlowComponents } from '../store/helpers';
 export class Assets {
     private endpoint: string;
     private localStorage: boolean;
-    protected items: SearchResult[] = [];
     protected nameProperty: string;
     protected idProperty: string;
 
+    // TODO: consider redux backed by local storage for this
+    protected items: SearchResult[] = [];
```
ycleptkellan: Is `objects` only going to be used for flows?

src/services/AssetService.ts
```diff
@@ -6,17 +7,43 @@ import { FlowComponents } from '../store/helpers';
 export class Assets {
     private endpoint: string;
     private localStorage: boolean;
-    protected items: SearchResult[] = [];
     protected nameProperty: string;
     protected idProperty: string;
 
+    // TODO: consider redux backed by local storage for this
```
ycleptkellan: Sorry, can you clarify the consideration here? All of this could exist in redux with checks to determine when to persist/hydrate asset state, right?

ycleptkellan: Nice, perhaps asset persistence is gated by an env variable?

ericnewcomer: The more I think about this the more I think we should remove the redux requirement for all of this since it isn't a necessary piece in the embedded case. If nothing is rendering based on it, I don't see the value in putting it in redux and it should probably be a goal that nothing is rendering off the assets from the `AssetService` since they are remote in the embed scenario. If that's the case then it would look more like:

#### Local Dev
 * Faux json asset endpoints
 * Assets merged with items stored in browser local storage

#### Netlify preview
 * Faux json asset endpoints
 * Assets merged with items encountered in loading of flow definitions stored in memory

#### Embed
 * All assets fetched on demand from asset endpoints, no local cache or storage

This means removing `flows` and `groups` from redux and fixing the dependencies on those to be on demand via the `AssetService`.

ericnewcomer: That said, local dev and netlify case might be implemented the same with the only difference being that the local dev case would just hydrate from local storage into memory and write it out as needed.

src/services/AssetService.ts
ericnewcomer: admittedly terrible names for both to these... will come up with something better as I move forward on this.

PR #288, opened 2018-04-26
Asset refactor

ericnewcomer:
* Remove groups, contacts, and flows from redux
* Remove SearchResult in favor of consistent use of Asset
* Uniform item storage in AssetService

Next Step: 
This temporarily disables FlowList. The next step will be to reimplement that as a simple control outside the editor backed by the AssetService (can be it's own).

### DIFF
--- a/src/services/AssetService.ts
+++ b/src/services/AssetService.ts
@@ -1,18 +1,25 @@
 // tslint:disable:max-classes-per-file
-import { SearchResult } from '../store';
 import { Endpoints, FlowEditorConfig, ContactProperties, AttributeType } from '../flowTypes';
 import axios, { AxiosResponse } from 'axios';
 import { FlowComponents } from '../store/helpers';
+import { dump } from '../utils';
+
+export interface Asset {
+    id: string;
+    name: string;
+    type: string;
+
+    isNew?: boolean;
+    content?: any;
+}
 
 export class Assets {
     private endpoint: string;
     private localStorage: boolean;
     protected nameProperty: string;
     protected idProperty: string;
-
-    // TODO: consider redux backed by local storage for this
-    protected items: SearchResult[] = [];
-    protected objects = {};
+    protected assetType: string;
+    protected assets: { [id: string]: Asset } = {};
 
     constructor(endpoint: string, localStorage: boolean) {
         this.localStorage = localStorage;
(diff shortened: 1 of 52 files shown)


src/services/AssetService.ts
```diff
@@ -1,18 +1,25 @@
...
+    type: string;
+
+    isNew?: boolean;
+    content?: any;
+}
 
 export class Assets {
     private endpoint: string;
     private localStorage: boolean;
     protected nameProperty: string;
```
ycleptkellan: should we be typing `nameProperty`, `idProperty` and `assetType`? we've used Enums in the past, curious whether you think thats good practice or not

src/services/AssetService.ts
```diff
@@ -1,18 +1,25 @@
 // tslint:disable:max-classes-per-file
-import { SearchResult } from '../store';
 import { Endpoints, FlowEditorConfig, ContactProperties, AttributeType } from '../flowTypes';
 import axios, { AxiosResponse } from 'axios';
 import { FlowComponents } from '../store/helpers';
+import { dump } from '../utils';
+
+export interface Asset {
```
ycleptkellan: just a head's up, `Simulator` has an `Asset` interface as well

src/services/AssetService.ts
ericnewcomer: agreed, enum makes more sense here

src/services/AssetService.ts
ericnewcomer: not for long it doesn't! simulator is going to need a bit of reworking for assetage as well, imagine it won't need it's own thing.

ericnewcomer:
Coverage dropped due to some stuff I needed to comment out, I'll make good in the next PR, I promise!

src/services/AssetService.ts
```diff
@@ -21,34 +28,49 @@ export class Assets {
...
+            const url = `${this.endpoint}?uuid=${id}`;
             axios
                 .get(url)
                 .then((response: AxiosResponse) => {
                     const ob = response.data.results[0];
+                    const asset = {
+                        id: ob.uuid,
+                        name: ob.name,
+                        type: this.assetType,
+                        content: ob.definition
```
ycleptkellan: just curious, why might a flow asset also need its definition?

ericnewcomer: this is the case of fetching by uuid and getting the full definition

ycleptkellan: ah, doh, thanks

PR #290, opened 2018-04-27
Use asset service for simulation

ericnewcomer:
woot woot.. next step is pointing our endpoints at a locally running asset server and making that work (ie, RP)

### DIFF
--- a/src/component/Simulator/Simulator.tsx
+++ b/src/component/Simulator/Simulator.tsx
@@ -90,12 +90,6 @@ interface Session {
     contact: Contact;
 }
 
-interface Asset {
-    type: string;
-    url: string;
-    content: any;
-}
-
 /**
  * Our dev console for simulating or testing expressions
  */
--- a/src/component/Simulator/Simulator.tsx
+++ b/src/component/Simulator/Simulator.tsx
@@ -226,8 +226,10 @@ export class Simulator extends React.Component<SimulatorProps, SimulatorState> {
         renderNodes.map((renderNode: RenderNode) => {
             nodes.push(renderNode.node);
         });
-        this.props.definition.nodes = nodes;
-        return this.props.definition;
+        return {
+            ...this.props.definition,
+            nodes
+        };
     }
 
     private startFlow(): void {
--- a/src/component/Simulator/Simulator.tsx
+++ b/src/component/Simulator/Simulator.tsx
@@ -20,6 +20,7 @@ import { RenderNodeMap, RenderNode } from '../../store/flowContext';
 import { getOrderedNodes } from '../../store/helpers';
 import { dump } from '../../utils';
 import { fakePropType } from '../../config/ConfigProvider';
+import AssetService, { getBaseURL } from '../../services/AssetService';
 
 const ACTIVE = 'A';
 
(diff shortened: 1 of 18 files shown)


src/services/AssetService.ts
```diff
@@ -135,6 +157,65 @@ export class Assets {
...
+                    })
+                    .catch(error => reject(error));
+            }
+        });
+
+        return result;
+    }
+
+    public getAssetSet(): any[] {
+        const assets: any[] = [];
```
ycleptkellan: small deal, but i think both `getAssetSet` and `getAssetContents` we should be making use of `Array.prototype.reduce` instead of a `map/push` combination. what do you think?

src/component/Simulator/Simulator.tsx
```diff
@@ -220,82 +220,28 @@ export class Simulator extends React.Component<SimulatorProps, SimulatorState> {
...
-                        uuid: '57f1078f-88aa-46f4-a59a-948a5739c03d',
-                        name: 'Simulator',
-                        address: '+12345671111',
-                        schemes: ['tel'],
-                        roles: ['send', 'receive']
-                    }
-                ]
-            }
-        ];
+        return this.props.definition;
```
ycleptkellan: don't know if it matters that we're reassigning props or if it could have any negative consequences but we could also return a new object, e.g. 

```
private getCurrentDefinition(): FlowDefinition {
    const renderNodes = getOrderedNodes(this.props.nodes);
    const nodes = renderNodes.map(({ node }: RenderNode) => node);
    return {
        ...this.props.definition,
        nodes
    };
}
```

src/services/AssetService.ts
ericnewcomer: Not sure how that would be nicer operating against a map. I think would agree if we could use `entries()` but since it's an array of keys, don't think reduce makes sense.

src/component/Simulator/Simulator.tsx
ericnewcomer: ya, this definitely felt hacky. New object coming up.

src/services/AssetService.ts
ericnewcomer: Fair enough though. Not wild about this pattern either so I’ll play with it a bit first.

ycleptkellan: Yeah not a big deal, FWIW was thinking:

```
return Object.keys(this.assets).reduce((assets, key) => {
    const { type, id, name, content } = this.assets[key];
    if (type !== AssetType.Property) {
        return [...assets, {
                [this.idProperty]: id,
                name,
                ...content
            }
		];
    }
    return assets;
}, []);
```

I think in the future a `forEach` would make more sense than `map` since `map` implies that a new array is being created.

ericnewcomer: Right. That’s a lot of dots. :) Lemme play with it a bit. I’m heading out right now but can look later.

ycleptkellan: Yeah yeah...😉 

`concat` would probably work as an alternative to array destructuring, i'm cool w/ a `forEach`/`push` combo tho 👍

ericnewcomer: Switched this to use `map` properly (instead of `forEach`), which I believe is the correct functional way to do this.

ycleptkellan: i think `reduce` is the ideal method here because it allows you to conditionally return a value to the accumulator. the problem w/ map is that it returns a value no matter what.

ycleptkellan: sounds good, no strong feelings either way re: `reduce` vs `forEach`/`push`. also, my b, hadn't seen that the new implementation of `getAssetContents` doesn't have an `if` block; in that case, agree, `map` is the method.

PR #291, opened 2018-04-30
Adjustments to prep for RP asset format

ericnewcomer:
Lots of changes to prep for RP integration. Adds ability to route to a local RP server via webpack dev server based on environment variables.


ericnewcomer:
netlify lambdas broken, not sure what's up there but will have a look tomorrow

PR #309, opened 2018-05-07
Apply isOptionUnique callback to GroupsElement

ycleptkellan:
- Solves the issue we were running into w/ `AttribElement` whereby whitespace wasn't 
being stripped from the value passed to the `createNewOption` handler.

- Extracts common handlers `isOptionUnique`, `isValidNewOption`, and `optionExists` out to utils.


ycleptkellan:
this one's ready as well

PR #312, opened 2018-05-09
More tests, shared node rendering

ericnewcomer:
* Also brings in new icon set


ericnewcomer:
Added bits to make sure we show the asset icons for all select search based widgets

PR #313, opened 2018-05-09
Set contact field, name refactor

ycleptkellan:
- Removes `set_contact_property` type, replaces it w/ `set_contact_name`. `set_contact_language`, `set_contact_timezone` to follow. 
- Pushes both actions' form state to the 'dux.


src/store/thunks.ts
```diff
@@ -497,8 +498,6 @@ export const onUpdateAction = (action: AnyAction) => (
         flowContext: { nodes }
     } = getState();
 
-    timeStart('onUpdateAction');
```
ericnewcomer: ?

PR #315, opened 2018-05-10
Small rendering fixes

ericnewcomer:
* Fix stray 'x' on single select widgets
* Fix missing styling on start flow action
* Render start flow with shared asset list renderer
* Fix switching to a split when there is no previous node

### DIFF
--- a/src/component/SelectSearch/SelectSearch.tsx
+++ b/src/component/SelectSearch/SelectSearch.tsx
@@ -249,7 +249,7 @@ export default class SelectSearch extends React.PureComponent<
                     valueKey="id"
                     labelKey="name"
                     multi={this.props.multi}
-                    clearable={this.props.multi}
+                    clearable={false}
                     searchable={true}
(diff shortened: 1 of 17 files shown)


ericnewcomer:
@ycleptkellan while adding coverage for SelectSearch to appease the coverage gods, I feel better about the onChange interface passed down to always be an array of selected items

src/component/SelectSearch/SelectSearch.tsx
```diff
@@ -66,45 +66,18 @@ export default class SelectSearch extends React.PureComponent<
...
-                this.props.onChange(selection);
-            }
-
-            this.setState(
-                {
-                    selections
-                },
-                () => this.select.focus()
-            );
-        }
```
ericnewcomer: instead have a shared implementation with the multi case

ycleptkellan: Oh sweet, that's a win for sure.

PR #317, opened 2018-05-10
Clear form state when done with node editor

ericnewcomer:


ericnewcomer:
Fixed this in the start_session PR

PR #323, opened 2018-05-11
Set Language Action

### LINKED ISSUE #107: Set language action
* An action to set the language for the current contact

ycleptkellan:
resolves #107 

- [x] Environment endpoint
- [x] Fetch languages on flow initialization, add them to ‘dux
- [x] Remove languages from context
- [x] Form state
- [x] Tests


ycleptkellan:
Note the 'Clear Value' option that appears in the language dropdown when a language has been added to the action. Should we allow users to clear that value?

src/utils/languageMap.ts
```diff
@@ -0,0 +1,742 @@
+import { Language } from '../flowTypes';
+
+export const languageMap: { [key: string]: Language } = {
```
ericnewcomer: would like to avoid having this exhaustive list here, looks like this is being used for the base language only being an iso code. we can look into options for this later tho.

src/component/actions/SetContactAttrib/SetContactAttribFormHelper.ts
```diff
@@ -70,6 +90,15 @@ export class SetContactAttribFormHelper implements FormHelper {
...
                     name: formState.value.value
                 } as SetContactName;
+            case Types.set_contact_language:
+                return {
+                    ...action,
+                    // we return an empty string to indicate the value is being cleared
+                    language:
+                        (formState as SetContactLanguageFormState).value.value.id === removeAsset.id
+                            ? ''
+                            : (formState as SetContactLanguageFormState).value.value.id
```
ericnewcomer: this blows up if user just accepts the form without first making a selection

src/services/AssetService.ts
```diff
@@ -75,7 +80,7 @@ export class Assets {
         this.idProperty = IdProperty.UUID;
     }
 
-    public get(id: string): Promise<Asset> {
+    public get(id: string = ''): Promise<Asset> {
```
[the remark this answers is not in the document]
ycleptkellan: I made this optional to allow me to query for the `environment` asset on line 176 in `thunks.ts`, but I see that's not actually necessary.

src/services/AssetService.ts
```diff
@@ -126,15 +131,28 @@ export class Assets {
...
 
         return axios.get(url).then((response: AxiosResponse) => {
-            for (const result of response.data) {
-                if (this.matches(term, result.name)) {
-                    matches.push({
-                        name: result.name,
-                        id: result[this.idProperty],
-                        type: this.assetType
-                    });
+            // Only attempt to match if response contains a list of externally-fetched assets
```
ycleptkellan: This is my answer to handling the single-asset response our mocked `environment` endpoint is currently returning, what do you think? This change was necessary after updating `SetContactAttribForm` to pass the `EnvironmentAsset` service to `SelectSearch`.

src/utils/languageMap.ts
ycleptkellan: Same, yeah this is just a first pass at getting the name of a language from its iso since that's all the `set_contact_language` action and `environment` response give us RN.

src/services/AssetService.ts
ycleptkellan: @ericnewcomer this one's ready for another review 👍

Agree that having duplicate `typeConfig`s and a language map aren't ideal, happy to address these down the road.

ericnewcomer:
looks like environment is empty on netlify preview. is it just them?

ycleptkellan:
@ericnewcomer sorry, should be fixed now

Issue #105, opened 2018-01-09
Add functions to completion


ericnewcomer:
* Auto complete functions
* Make completion work inside ()

PR #328, opened 2018-05-16
WIP: Result Count Updates

ycleptkellan:
- Changes `getSuggestedResultName` to only count nodes w/ msg waits.
- Changes `Response` to `Result`, as passive and hybrid splits aren't operating on a response.


ycleptkellan:
Ah, dang, thanks for catching that! Riffing on this in #331, I'll let you know when it's ready for review.

ycleptkellan:
@ericnewcomer this one's tricky in that we aren't yet maintaining a list of result names in memory (which, as you know, we'll have to do for the completion menu). Figure we should discuss expectations around that before moving forward on this. In short, no easy way to cover that edge case by my estimation.

ericnewcomer:
Fine punting on this edge case for now

ericnewcomer:
...so long as you file a ticket so we don't forget about it.

ycleptkellan:
I can bite this off in a two-part PR next week if you'll be refactoring router forms. Will leave this open until we have a chance to discuss.

ycleptkellan:
OK @ericnewcomer this one's ready for another pass. 

Once we've got this where we want it I figure one or more of the following make sense:

- Get #334 fixed so we can add logic to ensure we're removing `resultNames` from the store when the user changes a router to action.
- #104 
- #105. 

LMK what you think 👍

ycleptkellan:
Ah, thanks. Looking into it.

ericnewcomer:
@ycleptkellan this ready for another go?

ycleptkellan:
No quite yet, one last thing from that merge to fix.

PR #336, opened 2018-05-30
Form initialization

ericnewcomer:
Move towards a more general form initialization so router forms can follow the same pattern. Move localization form initialization into the form helpers.

* Remove localizations from flow context
* Add localizations to node editor settings
* Use node editor settings to initialize forms


src/component/actions/SendBroadcast/SendBroadcastFormHelper.tsx
```diff
@@ -1,12 +1,29 @@
...
+                    action = settings.localizations[0].getObject() as BroadcastMsg;
+                } else {
+                    return {
+                        type: Types.send_broadcast,
+                        text: { value: '' },
+                        recipients: { value: [] },
+                        valid: true
+                    };
+                }
+            }
```
ericnewcomer: This is an example of how our form helpers are going to be much more capable in this model.. this feels like the right place for this sort of thing.

src/component/actions/SetContactAttrib/SetContactAttribFormHelper.test.ts
```diff
@@ -18,51 +21,70 @@ const setContactNameAction = createSetContactNameAction();
 const setContactLanguageAction = createSetContactLanguageAction();
 const setContactChannelAction = createSetContactChannelAction();
 
-const setContactFieldFormState = formHelper.actionToState(
-    setContactFieldAction,
+const setContactFieldFormState = formHelper.initializeForm(
+    {
+        originalNode: null,
```
ericnewcomer: tests are the only place where originalNode can be null, this is why the data model requires it

src/store/flowContext.ts
```diff
@@ -137,18 +125,6 @@ export const dependencies = (
...
-    action: ActionTypes
-) => {
-    switch (action.type) {
-        case Constants.UPDATE_LOCALIZATIONS:
-            return action.payload.localizations;
-        default:
-            return state;
-    }
-};
-
```
ericnewcomer: die redux boilerplate, die!

ericnewcomer:
@ycleptkellan this is ready for your stamp

PR #342, opened 2018-05-31
Fix #334

### LINKED ISSUE #334: It isn't possible to change an existing router node to an action, save.
This throws the `Cannot find action...` error in `store/helpers.ts`' `getActionIndex` method.

ycleptkellan:
- Adds a mutator to replace a router node w/ an action, à la `spliceInRouter`. 
- Resolves #334.


ycleptkellan:
ready for ya @ericnewcomer

src/store/__snapshots__/thunks.test.ts.snap
```diff
@@ -1,443 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Flow Manipulation init should fetch and initalize flow 1`] = `
```
ycleptkellan: not quite sure what's going on here

src/store/mutators.ts
```diff
@@ -161,6 +161,31 @@ export const updateAction = (nodes: RenderNodeMap, nodeUUID: string, action: Any
     });
 };
 
+export const spliceInAction = (
```
ericnewcomer: feel like this closer to `convertToAction(..)`

src/store/mutators.ts
```diff
@@ -161,6 +161,31 @@ export const updateAction = (nodes: RenderNodeMap, nodeUUID: string, action: Any
...
+): RenderNodeMap => {
+    const { [nodeUUID]: originalRenderNode } = nodes;
+    return mergeNode(
+        nodes,
+        mutate(originalRenderNode, {
+            node: {
+                // Append action to node
+                actions: { $set: [action] },
+                // Off any exit but the first, remove the first's name
+                exits: { $set: [{ ...originalRenderNode.node.exits[0], name: null }] },
```
ericnewcomer: _Other_ exit instead like we discussed

src/store/mutators.ts
ycleptkellan: It's merging it into the `RenderNodeMap` as well though. Should this method only be concerned w/ converting the node to an action?

PR #344, opened 2018-06-01
Fix #343

### LINKED ISSUE #343: Editor throws when user changes the attribute in `SetContactAttribForm`

ycleptkellan:
Resolves #343 and another issue I ran into whereby `getActionIndex` is passed the `uuid` of the new action, which doesn't yet exist. You can reproduce by clicking on the first action, changing it to another action type and saving it. 

Need this for #104.

### DIFF
--- a/src/store/mutators.ts
+++ b/src/store/mutators.ts
@@ -152,14 +152,19 @@ export const addAction = (
  * @param nodeUUID
  * @param action
  */
-export const updateAction = (nodes: RenderNodeMap, nodeUUID: string, action: AnyAction) => {
+export const updateAction = (
+    nodes: RenderNodeMap,
+    nodeUUID: string,
+    newAction: AnyAction,
+    originalAction?: AnyAction
+) => {
     const nodeToEdit = getNode(nodes, nodeUUID);
     // If we have existing actions, find our action and update it
-    const actionIdx = getActionIndex(nodeToEdit.node, action.uuid);
+    const actionIdx = originalAction ? getActionIndex(nodeToEdit.node, originalAction.uuid) : 0;
     return mutate(nodes, {
         [nodeUUID]: {
             node: {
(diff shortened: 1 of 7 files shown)


src/store/mutators.ts
```diff
@@ -152,14 +152,19 @@ export const addAction = (
...
+export const updateAction = (
+    nodes: RenderNodeMap,
+    nodeUUID: string,
+    newAction: AnyAction,
+    originalAction?: AnyAction
+) => {
     const nodeToEdit = getNode(nodes, nodeUUID);
     // If we have existing actions, find our action and update it
-    const actionIdx = getActionIndex(nodeToEdit.node, action.uuid);
+    const actionIdx = originalAction ? getActionIndex(nodeToEdit.node, originalAction.uuid) : 0;
```
ycleptkellan: Lean on `originalAction` to determine index to update.

src/store/thunks.ts
```diff
@@ -526,13 +526,11 @@ export const handleTypeConfigChange = (typeConfig: Type, settings: NodeEditorSet
...
     if (typeConfig.formHelper) {
         // only use the original action if it is the same type
-        if (settings) {
-            settings.originalAction =
-                settings.originalAction && settings.originalAction.type === typeConfig.type
-                    ? settings.originalAction
-                    : null;
-        }
-        dispatch(updateForm(typeConfig.formHelper.initializeForm(settings, typeConfig.type)));
+        const customSettings =
```
ycleptkellan: We now expect settings to be passed to `handleTypeConfigChange`.

src/store/mutators.ts
ericnewcomer: node hear is originalNode to isn't it? seems should take that settings object or at least have consistent naming.

PR #347, opened 2018-06-04
Wire SetRunResult action to Redux Store

ycleptkellan:


ycleptkellan:
Next step in this direction would be invalidating nodes that reference deleted results. Should we focus on that next or couch it and move forward w/ the rest of the completion stuff (adding contact fields and functions)?

ericnewcomer:
I'd stay the course on completion before looking at deleted things, thats a different class of issues.

Issue #429, opened 2019-02-02
Creation of assets inside asset selectors


ericnewcomer:
We need to create assets on demand and also give feedback and block saving if they cannot be created (ie, for RP if they are over the limit for that type, etc).

 - [x] Groups
 - [x] Labels
 - [x] Contact Fields

PR #437, opened 2019-02-05
Create groups on the fly, switch to API v2

ericnewcomer:
* Use RP groups api endpoint to fetch / create groups
 * Honor CSRF cookie as http header
 * Fancy styling to make creation errors appear embedded within the control
 * Filter out dynamic groups from add/remove group actions


ericnewcomer:
#429

PR #446, opened 2019-02-22
AddURN action

ericnewcomer:


nicpottier:
Camelcase Urn is making my eye bleed! 😭

ericnewcomer:
So you prefer the AddURN convention I take it? That’s where I started but this is the first acronym I added and felt weird in places. I agree that probably feels overall more natural to me too though.

nicpottier:
Ya, doesn't really matter. Probably just the bias from the Go lang where all acronyms are all caps. But I think I do prefer it.

ericnewcomer:
Pretty sure I do too. Convinced, will change it.

PR #605, opened 2019-06-04
🗞️ Updates to auto-completion to reflect current engine context

rowanseymour:


src/components/form/textinput/TextInputElement.tsx
```diff
@@ -110,16 +110,16 @@ export class TextInputElement extends React.Component<TextInputProps, TextInputS
...
-      'date',
+      'child',
       'fields',
+      'input',
+      'parent',
+      'results',
+      'run',
+      'trigger',
       'urns',
-      'input'
```
rowanseymour: adding webhook and sorting a-z so it's easier to compare with goflow's list

rowanseymour:
@ericnewcomer turns out I don't quite understand how the snapshot tests work and how to update them etc...

ericnewcomer:
I'll take this and also spend some time cleaning up the README

rowanseymour:
@ericnewcomer won't be insulted if it's easier to start from scratch. @nicpottier and I chatted a bit about this at lunch and the end goal should be goflow generating a JSON map of the context from comments on code, and the editor consuming that.. but that's probably for another day.

ericnewcomer:
Totally agree on this being json!

PR #670, opened 2019-07-09
Fix operand for field delimiter

ericnewcomer:
This fixes the operand that is created as part of a delimited result option. Convert test to RTL and add delimiter test.


src/components/flow/routers/result/helpers.ts
```diff
@@ -110,15 +109,9 @@ export const stateToNode = (
...
+
   const asset = state.result.value;
-  if (asset.type === AssetType.URN) {
-    operand = `@(format_urn(contact.urns.${asset.id}))`;
-  } else if (asset.type === AssetType.Field) {
-    operand = `@contact.fields.${asset.id}`;
-  } else {
-    operand = `@contact.${asset.id}`;
-  }
+  const operand = `@(field(results.${asset.id}, ${state.fieldNumber}, '${state.delimiter}'))`;
```
rowanseymour: this isn't quite right - string literals in excellent are always with `"` @ericnewcomer 

sorry I should have actually reviewed this!

Issue #650, opened 2019-06-21
Completions from static JSON file


rowanseymour:
This would step 1 for https://github.com/rapidpro/rapidpro/issues/974, step 2 would be getting goflow to generate it from docstrings.

Started playing around with a possible schema:

```json
[
    {   
        "name": "contact",
        "description": "The current contact in the flow",
        "properties": [
            {
                "name": "name",
                "description": "The full name of the contact"
            },
            {
                "name": "language",
                "description": "The language code of the contact"
            },
            {
                "name": "channel",
                "description": "The preferred channel of the contact",
                "properties": [
                    {
                        "name": "name",
                        "description": "The name of the channel"
                    },
                    {
                        "name": "address",
                        "description": "The address of the channel"
                    }
                ] 
            },
            {
                "name": "fields",
                "description": "The contact's custom fields",
                "dynamic_properties": {
                    "source": "field_keys",
                    "template": {
                        "description": "{property.name} for the contact"
                    }
                }
            }
        ]
    },
    {
        "name": "results",
        "description": "The current results in this run",
        "dynamic_properties": {
            "source": "result_keys",
            "template": {
                "description": "{property.name} for the run",
                "properties": [
                    {
                        "name": "value",
                        "description": "The value of {property.name}"
                    }, 
                    {
                        "name": "category",
                        "description": "The category of {property.name}"
                    }   
                ]
            }
        }
    },
    {
        "name": "groups",
        "description": "The groups for the contact",
        "array_properties": {
            "template": {
                "description": "{property.name} for the run",
                "properties": [
                    {
                        "name": "uuid",
                        "description": "The UUID of group {property.index}"
                    }, 
                    {
                        "name": "name",
                        "description": "The name of group {property.index}"
                    } 
                ]
            }
        }
    }
]
```

rowanseymour:
Some thoughts from playing with this on the goflow side:

- It's would be a lot easier to have a single set of descriptions for a single type, i.e. the description for a field of a contact is the same for `run.contact`, `child.contact` or `parent.contact`. Same for a channel whether it's `contact.channel` or `input.channel`. I think this is ok because it should be obvious to the user from the context path which contact they are referencing. For example in this case, the description would be just _The first name of the contact_.

<img width="329" alt="Captura de Pantalla 2019-06-25 a la(s) 10 39 25" src="https://user-images.githubusercontent.com/675558/60112524-9166b900-9735-11e9-8bd3-665fdb35b976.png">

[...]

PR #678, opened 2019-07-18
Remote Completion

ericnewcomer:
Replace completion logic with remote schema.

Fixes: #650


rowanseymour:
@ericnewcomer so the general structure works ok? I was thinking after I made it it would be easier if it included all types including primitive types - that way for every `"type": "xx"` there is a corresponding type def `{"name": "xx", ...}`. But your call since you're the one consuming this.

ericnewcomer:
Ya, I think it's okay as is. That said, adding types would be nice if we ever wanted to surface them with iconography.

PR #703, opened 2019-09-24
Add contact_query to start session form

ericnewcomer:


src/components/flow/actions/sendbroadcast/SendBroadcast.module.scss
```diff
@@ -28,7 +28,6 @@
   .message {
     background: #fff;
     padding: $action_padding;
-    width: 100%;
```
ericnewcomer: Not related, just notice this style was breaking wrapping

Issue #701, opened 2019-09-23
Delimited field split should use zero-based index


rowanseymour:
If you pick first item, generated operand is...

```
@(field(results.resp1, 1, \".\"))
```

ericnewcomer:
I swear excellent was one-based for these at some point. Maybe that was the old excellent? Will change.

rowanseymour:
Excellent 1.0 aka Classic Edition was always 1-based like Excel

ericnewcomer:
Classic Edition or Excellent 2017?

PR #713, opened 2019-09-27
Zero based fields on delimited splits

ericnewcomer:
Fixes: #701


ericnewcomer:
One side-effect here is that we are now showing what the real operand is, so if they had `first` before (we were dealing with it as second), we now show `second`. Behavior in the flow will remain the same.

nicpottier:
I think that's right.

Issue #700, opened 2019-09-18
Plural for continue for no response wrong


nicpottier:
![Screen Shot 2019-09-18 at 9 44 35 AM](https://user-images.githubusercontent.com/168067/65168275-037b3100-d9f9-11e9-8f45-8299076a4722.png)

PR #716, opened 2019-09-27
Fix timeout plural for one day

ericnewcomer:
Fixes: #700


ericnewcomer:
That's a funny sounding PR name, this fixes it forever though.

PR #718, opened 2019-09-30
Switch media operand to @input

ericnewcomer:


ericnewcomer:
That's correct, will affect gps, etc as well.

src/components/nodeeditor/constants.ts
```diff
@@ -2,7 +2,7 @@ export const DEFAULT_OPERAND = '@input.text';
 export const GROUPS_OPERAND = '@contact.groups';
 export const WEBHOOK_OPERAND = '@run.webhook';
 export const SUBFLOW_OPERAND = '@child.run.status';
-export const MEDIA_OPERAND = '@input.attachments';
+export const MEDIA_OPERAND = '@input';
```
rowanseymour: We just to ensure this matches https://github.com/nyaruka/goflow/blob/master/legacy/definition.go#L761 (or vice versa)

Does media include waiting for digits in IVR flows?

ericnewcomer: media does not include digits, but this was using `@input`. Changing it to `@input.text` to match.

PR #724, opened 2019-10-19
Split by intent

ericnewcomer:
Add support for classifiers. This should now be functional. Only other thing would be maybe beta gating and adding coverage.


nicpottier:
Looks good. One small thing is maybe make "categorize as" nowrap?

Screen Shot 2019-10-21 at 7 22 00 AM

So is this using `_` result names? Saw a comment there but couldn't quite tell what was happening.

ericnewcomer:
The categorize as wrapping doesn't occur when embedded in RP. Planned to determine what style is conflicting there to fix the netlify preview instead. But planned to do this after merging.

What do you mean by saw a comment? In the product? They shouldn't see it surfaced anywhere, but yes, it's using underscore for the classification result.

nicpottier:
Sorry, wasn't clear was just talking about reading through the code, maybe it wasn't a comment. Sounds good on the wrapping.

PR #726, opened 2019-10-21
Update airtime node generation to use has_category test

rowanseymour:


src/components/flow/routers/airtime/__snapshots__/AirtimeRouterForm.test.ts.snap
```diff
@@ -143,7 +143,7 @@ Array [
               "Success",
             ],
             "category_uuid": "763e4844-3e1b-407a-a1b5-5fdfcd308b41",
-            "type": "has_only_text",
+            "type": "has_category",
```
rowanseymour: this is still coming back as `has_only_text`.. and I'm very confused as to why..

rowanseymour: n/m realized tests use a different function to create this part..

Issue #71, opened 2017-11-27
Localization


ycleptkellan:
[React Int'l](https://github.com/yahoo/react-intl) is commonly used for this sort of thing.

PR #741, opened 2019-10-31
i18n support on FlowEditor [blank pr]

johncordeiro:
Hey guys, I would like to propose an architecture for translation support on this project. I've checked some libraries to translate React components in a better way and the best one was [react-i18next](https://github.com/i18next/react-i18next). This architecture solves:

- Simple configuration for n-languages translation;
- Scanning translatable strings and generating locale files to translate;
- Translation inside or outside components;

Question for you:

- How to get the user-configured language and passing it to `i18n.js` file?


johncordeiro:
@ericnewcomer and @rowanseymour, any thoughts on this?

ericnewcomer:
i18next is interesting, did a bit of poking around and it's not super clear to me if this can be used with transifex which I see as a hard requirement for rapid projects. We need a full circle story with transifex to move forward with a localization strategy.

ericnewcomer:
A bit more detail.. it looks like transifex supports a variety of different JSON files. This looks like a basic key/value resource file which perhaps could work, but I don't see the original base language here for it to work off of.

ericnewcomer:
@johncordeiro, I looked into this a bunch today and I'm really not in love with any of the tooling around localization for javascript, but i18next seems as reasonable as the lot. On a separate branch I have something that I think will work for transifex, I'll open a PR with that shortly.

johncordeiro:
ok @ericnewcomer, let me know once you do that because I can help to make every string translatable.

ericnewcomer:
Hey @johncordeiro, opened #744. Did a bunch of research around making this approach work with transifex and I think I've massaged things into a manner that will work for us. See that PR for details. I already took the time to localize everything I could find. If you can spot any unlocalized strings (I'm sure there's at least a few), mention them over there and I'll make sure I include them in the first go.

Thanks for pushing me on this, it's been on my back burner for too long (Issue #71!)

PR #744, opened 2019-11-08
Add localization

ericnewcomer:
After reviewing various options `i18next` and it's friends `react-i18next` and `i18next-scanner` look to be as good an option as any (h/t @johncordeiro). However, after lots of fiddling, I went with keyed localization with fallbacks. What this means is that we won't break the localization connection if we tweak the wording or formatting of the base language. This still gives us the benefit of simple in-code fallbacks since the default is provided along with the key.

 * Create default.json with keys and fallbacks, generated with `% yarn scan`
 * Use structured json for the localization, I confirmed this matches the key/value json option for transifex and I tested full circle language updates using it.
 * Add en to override plurals as an example
 * Took several passes through the code to localize everything I could find that made sense, but I invariably missed something.

Fixes: #71


johncordeiro:
hey @ericnewcomer, good job man. Taking a look and sharing any comments with you soon. One question: have you created a project on Transifex for translation already? If yes, let me know so we can start translating.

ericnewcomer:
@johncordeiro, i've built the resource file and tested it, I will add it to transifex after merging this.

ericnewcomer:
Hey @johncordeiro any thoughts on this? Would like to get it merged. Thanks.

src/components/flow/actions/localization/KeyLocalizationForm.tsx
```diff
@@ -61,8 +62,11 @@ export default class KeyLocalizationForm extends React.Component<
 
   private getButtons(): ButtonSet {
```
johncordeiro: Is there a way to not duplicate this code in every component?

ericnewcomer: Fair point. Ya, this is a product composition over inheritance. It's harder to construct it with a helper since such a helper would need reference to the event handlers. But ya, there's a super common pattern for those dialogs, so maybe a helper function that takes a yay and a nay handler wouldn't be terrible.

ericnewcomer: For goflow docs, that is definitely trickier. Those are loaded from a file generated from goflow, so the localizations would ideally be created there. @rowanseymour, thoughts?

rowanseymour: @johncordeiro we plan to add that to goflow for the 5.4 release (general idea described at https://github.com/nyaruka/goflow/issues/799)

ericnewcomer:
@johncordeiro heads up that you can find the resource at https://www.transifex.com/rapidpro/flow-editor/ and is ready for you guys to add pt translations.

PR #752, opened 2019-11-19
Add twitter handle scheme back in

ericnewcomer:


src/components/flow/actions/addurn/AddURN.tsx
```diff
@@ -6,10 +6,14 @@ import i18n from 'config/i18n';
...
 export const MAX_TO_SHOW = 3;
 
-const AddURNComp: React.SFC<AddURN> = ({ scheme, path }): JSX.Element => (
-  <>
-    {i18n.t('add', 'Add')} {getSchemeObject(scheme).name} {path}
-  </>
-);
+const AddURNComp: React.SFC<AddURN> = ({ scheme, path }): JSX.Element => {
+  const schemeObject = getSchemeObject(scheme);
+  const schemeName = schemeObject ? schemeObject.name : scheme;
```
ericnewcomer: As a failsafe, fallback to the scheme as the name if we don't recognize it.

Issue #761, opened 2019-12-02
Localization of emails in simulator broken


rowanseymour:
<img width="266" alt="Captura de Pantalla 2019-12-02 a la(s) 15 48 32" src="https://user-images.githubusercontent.com/675558/69994131-35d9fc00-151b-11ea-92ed-e3e236169e2d.png">

PR #764, opened 2019-12-03
Disambiguate simulator email i18n key

ericnewcomer:
Fixes: #761

### DIFF
--- a/src/config/i18n/defaults.json
+++ b/src/config/i18n/defaults.json
@@ -135,6 +135,10 @@
   "create_group": "Create Group",
   "create_label": "Create Label",
   "drag_helper": "To connect nodes, <1>drag</1> from the red circle.",
+  "email": {
+    "subject": "Subject",
+    "to": "To"
+  },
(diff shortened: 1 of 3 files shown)


src/config/i18n/defaults.json
```diff
@@ -286,6 +286,7 @@
       "to": "To"
     },
     "sent_email_dialog": "Email Details",
+    "sent_email_summary": "Sent email to \"[[recipients]]\" with subject \"[[subject]]\"",
```
rowanseymour: so the structure of this arbitrary? like how subject is nested under sent_email?

ericnewcomer: Dot notation is used to nest things where appropriate. Sent email has the to and subject labels in different places so have those nested, since this is a single blurb, opted for a single un-nested key.

Alternative would be something like this which I don't hate:

```
sent_mail
  summary
  dialog
     title
     to
     subject
```

ericnewcomer:
Saw there were no translations yet, so reworked those keys a bit.

PR #770, opened 2019-12-17
Add facebook topic selection

ericnewcomer:


rowanseymour:
Maybe these channel specific things deserve to be to the right of advanced? No strong feelings

ericnewcomer:
Was toying with sharing a tab for them, problem is WhatsApp kind of needs the whole dialog. I suppose if we are open to internal scrolling (which relatedly is needed for the splits), then we could reconsider. I put it on it's own tab for now and tightened up the tab padding.

Issue #772, opened 2019-12-18
URL validation doesn't allow certain expressions


rowanseymour:
For example this URL can't be saved because of the `""`characters:

```
https://example.com?phone=@(url_encode(default(urn_parts(urns.tel).path, ""))
```

I'm not entirely sure validation makes much sense here since anything can be an expression... maybe skip validation when input contains `@`. For example this should be valid..

```
@contact.fields.update_url
```

nicpottier:
Ya since this can contain expressions there really isn't any validation whatsoever we can do here that makes much sense.

rowanseymour:
In fact some say in late 2019 it's best practice to write your webhooks like..

```
@globals.contact_tracker_webhook?phone=@urns.tel
```

🤔

PR #773, opened 2019-12-19
Add conditional validation for urls

ericnewcomer:
Fixes: #772


nicpottier:
So basically this is only doing URL validation in the cases where there are no expressions is that right?

rowanseymour:
Looks good - just need https://github.com/nyaruka/goflow/pull/815 as well so that we handle invalid URLs gracefully in the engine

Issue #786, opened 2020-01-15
Allow webhook bodies for all HTTP methods


rowanseymour:

rowanseymour:
Started working on this and realized it's trickier than it sounds because we want non-GETs to use the default but not GETs otherwise we'll start sending bodies with everyone's GET requests.

I think the key question is - if you have a POST request and you modify the body, and change it to a GET request, do we keep the [large] body or default it back to empty?

nicpottier:
Give that bodied GETs are pretty rare it seems like the more expected behavior would be to clear the body if you switch from POST to GET.

rowanseymour:
@ericnewcomer what does opt-in mean in this case? TBH I'm starting to feel like this is complicating the UI for a very edge casey feature

nicpottier:
Totally agree this is super edge case and happy to bag it, but can't we just change the "Post Body" tab to just be "Body" and always show it? It is empty for the GET case and gets cleared whenever you switch to GET from POST. When you switch to POST/PUT it is set to our default payload.

ericnewcomer:
That works for me.

rowanseymour:
I can do that - downside is just that someone potentially makes a bunch of changes to a POST payload and then loses them toggling to GET and back to POST.

rowanseymour:
Or maybe they intend to make a POST with a body so they edit the body, then change to POST, and they lose their body

nicpottier:
Yep, I'm comfortable with that for the time being.

chris-erickson:
Since I pushed for the feature, I would be completely OK with destroying your work switching between types.  Something people will learn quickly and I don't think people will be toggling this often.

PR #788, opened 2020-01-16
💁 Allow customization of body for GET webhooks

rowanseymour:
Adds #786

### DIFF
--- a/src/components/flow/routers/webhook/WebhookRouterForm.tsx
+++ b/src/components/flow/routers/webhook/WebhookRouterForm.tsx
@@ -79,14 +79,17 @@ export default class WebhookRouterForm extends React.Component<
     let ensureEmptyHeader = false;
 
     if (keys.hasOwnProperty('method')) {
-      const oldMethod = this.state.method.value.value;
-      const newMethod = keys.method.value;
       updates.method = { value: keys.method };
 
-      if (oldMethod === Methods.GET && newMethod !== Methods.GET) {
-        updates.body = { value: DEFAULT_BODY };
-      } else if (oldMethod !== Methods.GET && newMethod === Methods.GET) {
-        updates.body = { value: null };
+      const oldMethod = this.state.method.value.value;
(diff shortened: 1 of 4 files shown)


src/components/flow/routers/webhook/WebhookRouterForm.tsx
```diff
@@ -79,14 +79,14 @@ export default class WebhookRouterForm extends React.Component<
...
-      if (keys.method.value !== GET_METHOD.value) {
-        if (!this.state.postBody.value) {
-          updates.postBody = { value: DEFAULT_BODY };
-        }
-      } else {
-        updates.postBody = { value: null };
+      if (oldMethod === Methods.GET && newMethod !== Methods.GET) {
+        updates.body = { value: DEFAULT_BODY };
+      } else if (oldMethod !== Methods.GET && newMethod === Methods.GET) {
+        updates.body = { value: null };
```
ericnewcomer: This should probably be calling `getDefaultBody(method)` since that returns an empty string which will not match later equivalence checks.

src/components/flow/routers/webhook/WebhookRouterForm.tsx
```diff
@@ -236,49 +236,47 @@ export default class WebhookRouterForm extends React.Component<
...
+                body,
+                valid: this.state.valid && !hasErrors(body)
+              });
+            }}
+            autocomplete={true}
+            textarea={true}
+          />
+        </div>
+      ),
+      checked: this.state.body.value !== getDefaultBody(method)
```
ericnewcomer: This will resolve to true if you toggle to post and then back to GET without modifying the body (which should be unchecked)

src/components/flow/routers/webhook/WebhookRouterForm.tsx
rowanseymour: updated to always use `getDefaultBody`

Issue #793, opened 2020-01-21
Default webhook headers not right


ericnewcomer:
Should be setting `Accept` instead of `Content-Type` to `application/json`

rowanseymour:
Legacy flow migration sets `Content-Type` on the action if the method is `POST`.. we're also setting a default payload at that point which we know is JSON. I think we want to set both because a endpoint could need `Content-Type` set..

nicpottier:
For `POST`s, especially for legacy ones we should set `Content-Type` for sure. But with `Accept` as well for everything (by default)

ericnewcomer:
Yes, POSTs also need content type.

PR #795, opened 2020-01-22
More appropriate webhook header defaults

ericnewcomer:
Fixes: #793

### DIFF
--- a/src/components/flow/routers/webhook/WebhookRouterForm.tsx
+++ b/src/components/flow/routers/webhook/WebhookRouterForm.tsx
@@ -84,34 +84,34 @@ export default class WebhookRouterForm extends React.Component<
       const oldMethod = this.state.method.value.value;
       const newMethod = keys.method.value;
 
-      // if we don't have a content-type when switching to a PUT or POST, add one
-      if (oldMethod === Methods.GET && newMethod !== Methods.GET) {
-        const existing = this.state.headers.find(
+      if (oldMethod !== newMethod) {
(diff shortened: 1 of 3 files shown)


src/components/flow/routers/webhook/WebhookRouterForm.tsx
```diff
@@ -76,18 +76,41 @@ export default class WebhookRouterForm extends React.Component<
...
+              : null;
+          if (lastHeader && !lastHeader.value.name) {
+            uuid = lastHeader.value.uuid;
+          }
+          keys.header = { uuid, name: 'Content-Type', value: 'application/json' };
+        }
+        updates.body = { value: getDefaultBody(newMethod) };
+      } else if (oldMethod !== Methods.GET && newMethod === Methods.GET) {
+        // remove content type when resetting to a get
+        const existing = this.state.headers.find(
```
nicpottier: can this live outside the if to avoid the repetition?

ericnewcomer: it could but then it would be executed for every single change in that dialog instead of just method changes. could wrap the whole thing checking for method change which would allow that tho

nicpottier: Oh I see, well fine either way.

ericnewcomer: There was actually two bits of repetition, so i liked ur suggestion better

Issue #804, opened 2020-01-30
Simulator doesn't render contact_groups_changed events with both adds and removes


rowanseymour:
These can happen with dynamic groups in response to field changes

Issue #790, opened 2020-01-17
Simulator doesn't render input_labels_added events


rowanseymour:
And warning appears in console

PR #805, opened 2020-01-30
Simulator event rendering fixes

rowanseymour:
* Render input_labels_added events (#790)
* Fix rendering of contact_groups_changed events (#804)
* Adds tests!

### DIFF
--- a/src/components/simulator/__snapshots__/LogEvent.test.tsx.snap
+++ b/src/components/simulator/__snapshots__/LogEvent.test.tsx.snap
@@ -1,0 +1,324 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`LogEvent should render broadcast_created event 1`] = `
+<body>
+  <div>
+    <div
+      class="evt"
+    >
+      <div
+        class="send_msg"
+      >
+        <div
+          class="msg_text"
+        >
+          Hi there
+        </div>
+      </div>
+    </div>
+  </div>
+</body>
+`;
+
+exports[`LogEvent should render contact_field_changed event 1`] = `
+<body>
+  <div>
+    <div
+      class="evt"
+    >
+      <div
+        class="info"
+      >
+        <span>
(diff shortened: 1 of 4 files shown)


src/config/i18n/defaults.json
```diff
@@ -300,12 +300,13 @@
...
     "flow_entered": "Entered flow \"[[flow]]\"",
     "httplog_dialog": "HTTP Request Details",
+    "input_labels_added": "Message labeled with ",
     "msg_wait": "Waiting for reply",
     "removed_from_group": "Removed from ",
     "resthook_called": "Triggered flow event \"[[resthook]]\"",
     "run_result_changed": "Set result \"[[field]]\" to \"[[value]]\"",
     "sent_email": {
-      "summary": "Sent email to \"[[recipients]]\" with subject \"[[subject]]\"",
+      "summary": "Sent email to [[recipients]] with subject \"[[subject]]\"",
```
rowanseymour: Fixes whole list having quotes instead of each item in the list

src/components/simulator/__snapshots__/LogEvent.test.tsx.snap
```diff
@@ -0,0 +1,324 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
rowanseymour: big fan of snapshot testing now - it's the only sensible way to check HTML outputs

ericnewcomer: Totally agree. Also, while probably not a huge deal for the log event stuff, but as a general rule for snapshot testing, I try to have pertinent asserts along with the snapshot. Otherwise, it can be really easy to update a snapshot without realizing you are changing an expected truth.

Here is an example:
https://github.com/nyaruka/floweditor/pull/801/files#diff-05f981483ae9b76ba83005bfafd192b3R45

Basically, just treat the snapshot as a bonus assertion.

Issue #630, opened 2019-06-14
Add split by scheme?


nicpottier:
It is a fairly common thing to want to go down different paths based on the scheme of the contacts URN. This can be done via "Split by custom expression" but thats a high bar for people to pass. Would be nice for this to be a first order editor thing.

PR #809, opened 2020-02-05
Add split by scheme

ericnewcomer:
Fixes: #630 

In support of
https://github.com/rapidpro/rapidpro/issues/951


nicpottier:
So we are calling this "Split by Channel Type" to the user? Agree that may be more user friendly but man are we going to get confused about that?

nicpottier:
Maybe `Split by URN Type`? Have we accepted users need to know what a URN is? Maybe `Split by Address Type`?

Don't know, just slightly concerned about `Channel Type` since that definitely means something else in our world.

ericnewcomer:
Ya, I totally feel you. That said, I also feel like both of those suggestions are worse descriptors for the user as to what they are trying to do with this split though. In code, it is `split_by_scheme` and `SchemeRouterForm`. I'd rather we be confused than our users, but this was one of the things I wanted feedback on. Since we are exposing URN as a thing to users, an argument can be made there, but I bet you nearly 100% of our users that see that won't have a clue what that means. @rowanseymour ?

ericnewcomer:
I disagree but the nays have it! Will change.

ericnewcomer:
Switching to `Split by URN Type`

Summary text for the form:
> The contact's URN is the address they used to reach you such as a phone number or a Facebook ID. Select which URN types to split by below.

Thoughts?

nicpottier:
Ya, I think "URN Type" is exactly the right balance there as opposed to Scheme.

PR #811, opened 2020-02-06
Tweak operands used in split by contact field routers

### LINKED ISSUE #682: Tweak URN split operands
We'll get this right eventually but we #676 we changed URN type splits to use `@(urn_parts(urns.<scheme>).path)`... but now that `urn_parts` errors for non-URNs, we need to wrap this in `default()`.. thus:

`@(default(urn_parts(urns.<scheme>).path, ""))`

rowanseymour:
Fixes #682

### DIFF
--- a/src/components/flow/routers/field/helpers.ts
+++ b/src/components/flow/routers/field/helpers.ts
@@ -95,9 +95,9 @@ export const stateToNode = (
   let operand = DEFAULT_OPERAND;
   const asset = state.field.value;
   if (asset.type === AssetType.Scheme) {
-    operand = `@(urn_parts(urns.${asset.id}).path)`;
+    operand = `@(default(urn_parts(urns.${asset.id}).path, ""))`;
   } else if (asset.type === AssetType.Field) {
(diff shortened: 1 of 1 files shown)


src/components/flow/routers/field/helpers.ts
```diff
@@ -95,9 +95,9 @@ export const stateToNode = (
   let operand = DEFAULT_OPERAND;
   const asset = state.field.value;
   if (asset.type === AssetType.Scheme) {
-    operand = `@(urn_parts(urns.${asset.id}).path)`;
+    operand = `@(default(urn_parts(urns.${asset.id}).path, ""))`;
```
nicpottier: I'm confused here, the asset type says scheme but we are looking at path?

rowanseymour: this is when you're splitting on something like Phone Number - you select a scheme and then it's a split on the path of the URN with that scheme

rowanseymour:
@ericnewcomer less critical because not so common for `contact.urn` to be empty, but yeah should use default just in case

Issue #734, opened 2019-10-23
Missing dependencies


ericnewcomer:
Determine the strategy for missing dependencies when editing a flow. These can be the result of an import or a dependency being removed out from under them.

 - Groups
 - Contacts
 - URNs
 - Channels
 - Classifiers
 - Resthooks

Probably others. 

Current thinking is that it should not be an error to save a flow with missing dependencies, but that we should present the user with warnings throughout the interface so they are aware. Need to consider their use in both actions and routers, also how they are shown on the overview and how we deal with edit forms for each missing type.

rowanseymour:
The dependencies that goflow can enumerate using flow inspection are:

* Classifiers
* Channels
* Contacts
* Fields
* Flows
* Groups
* Labels
* Templates

And for all except contacts it can verify if the dependency is available. Obviously has no idea which contacts exist.

URNs are just strings and resthooks are only referenced by string so they've always been considered "soft".

ericnewcomer:
See: https://github.com/rapidpro/rapidpro/issues/1089

PR #814, opened 2020-02-09
Flow Issues

ericnewcomer:
Fixes: #734

Screen Shot 2020-02-20 at 2 07 51 PM
Screen Shot 2020-02-20 at 2 08 16 PM


ericnewcomer:
A few things to note:

1) Issues are presented with red babershop title
2) Issues are filtered by localization settings (so you may not see all issues without toggling for language)
3) Since our tomato red is co-opted for issues (or other warnings, deletion, etc), changed the Remove from Group action to be dark gray instead of red -- which always felt aggressive to me.  That said, this a departure from what users might be used to, but worth training them away from it now with the issue changes.
4) Missing dependencies are not surfaced specifically, only raised on the per-dialog issue listing

Still want to add a poptab widget for global issues, but that should be another PR

PR #837, opened 2020-03-04
Z indexes and you

ericnewcomer:
For anyone interested in what's going on here.. collecting my thoughts here for posterity.

Primary issue with z-indexes is that the flow nodes and the svg graphics for the plumbing have different parents. This of course limits the flexibility of z-index since they have distinct stacks. In order to have activity, recent messages, and connections above other nodes nodes they need to be descendants of a container within the svg tree.

Alas, this is possible, but it is a pretty big change. To accomplish this, Exit components need to use a "Portal" in React parlance which allows it to mount parts of itself somewhere else in the DOM. It all feels a bit crazy on the implementation, but in my tests seems to work well and gets all our bits layered quite nicely.

To make this happen, this PR also creates a jsPlumb "label" upon connection which creates a div within said connection svg. When that is connected, a callback makes its way back to the Exit requesting it and gives it a chance to attach its Portal to that element as described above. The upside of this means that our activity data is now genuinely attached to the path svg instead of faked like before. So, for example, if the path is shortened or gaps change on the stubs, they'll do the right thing.


nicpottier:
How do you feel about the inclusion of this for 5.4? Should this go into 6.0 and we just live with the stack order for 5.4?

ericnewcomer:
Was having the same internal dialogue. I feel like I'd like to merge it for 5.4, but only if we push it and start using it today.

PR #844, opened 2020-03-12
Switch TypeList to use  temba-select

ericnewcomer:
First use of temba-components in flow-editor, wooo.


ericnewcomer:
This needs searching support for already set selects to maintain functionality

PR #846, opened 2020-03-17
Saving Improvements

ericnewcomer:
Tries to address possibility of stale implicit saving and save lockout. Also adds a separate monitor to check for stale changes.

### DIFF
--- a/src/components/flow/Flow.tsx
+++ b/src/components/flow/Flow.tsx
@@ -39,7 +39,6 @@ import {
   UpdateSticky
 } from 'store/thunks';
 import {
-  ACTIVITY_INTERVAL,
   createUUID,
   isRealValue,
   NODE_PADDING,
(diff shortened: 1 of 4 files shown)


src/components/flow/Flow.tsx
```diff
@@ -191,15 +190,6 @@ export class Flow extends React.Component<FlowStoreProps, {}> {
...
 
-  public UNSAFE_componentWillUpdate(prevProps: FlowStoreProps): void {
-    if (
-      prevProps.editorState.activityInterval === this.props.editorState.activityInterval &&
-      this.props.editorState.activityInterval !== ACTIVITY_INTERVAL
-    ) {
-      this.props.mergeEditorState({ activityInterval: ACTIVITY_INTERVAL });
-    }
-  }
-
```
ericnewcomer: This is addressing a react warning. This is moved to the save method so that components can't update themselves after render.

src/config/i18n/defaults.json
```diff
@@ -237,8 +238,13 @@
     "type_label": "When a contact arrives at this point in your flow..."
   },
   "groups": "Groups",
+  "issue": {
+    "legacy_extra": "Invalid regular expression found: [[regex]]"
+  },
```
ericnewcomer: missed some i18n on issues

PR #848, opened 2020-03-18
Fix breaking out like-named routes

### LINKED ISSUE #847: Renaming repeated categories creates exits with duplicate UUIDs
https://app.intercom.com/a/apps/l6e7tr01/inbox/inbox/conversation/26207653563

Looks like you can recreate this on any ruleset where categories repeat, and then you try to change one of the repeated categories. For example given

Captura de Pantalla 2020-03-18 a la(s) 09 20 15

I can't change _Others_ unless I change both of them. Otherwise get validation failed dialog, and POST shows flow has duplicate exit UUIDs for this router.

ericnewcomer:
Fixes #847

### DIFF
--- a/src/testUtils/matchers.ts
+++ b/src/testUtils/matchers.ts
@@ -11,6 +11,27 @@ interface MatchResult {
   pass: boolean;
 }
 
+function toBeUnique<T>(this: jest.MatcherUtils, received: any[]): MatchResult {
+  const seen: { [value: string]: boolean } = {};
+
+  for (const item of received) {
+    const stringified = JSON.stringify(item, null, 1);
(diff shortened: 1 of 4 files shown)


src/testUtils/matchers.ts
```diff
@@ -11,6 +11,27 @@ interface MatchResult {
...
+    } else {
+      seen[stringified] = true;
+    }
+  }
+
+  return {
+    message: () => 'Array is unique',
+    pass: true
+  };
+}
```
ericnewcomer: Handy new jest matcher checking that the stringified value of each object in an array is unique

PR #878, opened 2020-05-04
🎟️ Tickets

rowanseymour:
Captura de Pantalla 2020-05-04 a la(s) 15 27 42

Captura de Pantalla 2020-05-04 a la(s) 15 29 46


src/components/flow/routers/ticket/TicketRouterForm.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    it('should cancel changes', () => {
+      const components = setup(true, { updateRouter: setMock() });
+      const instance = components.instance as TicketRouterForm;
+
+      instance.handleSubjectUpdate('New subject');
+      instance.getButtons().secondary.onClick();
+      expect(components.props.updateRouter).not.toBeCalled();
+    });
+  });
+});
```
ericnewcomer, line 43: I wish I had mentioned this earlier, but we've been phasing on `composeComponentTestUtils` for the excellent and much clearer react testing library. A more current router test example would be [WaitRouterForm.test.tsx](https://github.com/nyaruka/floweditor/blob/master/src/components/flow/routers/wait/WaitRouterForm.test.tsx)

Main difference here, is that accessing `instance` and calling functions directly is considered no-bueno. Instead, interact as a user firing events and then checking what happened.

rowanseymour: When I added this it did feel dirty to have to go back and make some methods in the component public just to call them here. Will update..

ericnewcomer: There's a ton more (and probably better) examples if you look at tests that are `test.tsx` files. Webhook and Random are a couple others. Some have a bit of both, been refactoring those a bit at a time when I touch them.

src/components/flow/routers/ticket/TicketRouterForm.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      // set our subject and result name
+      fireChangeText(subject, 'Need help');
+      fireChangeText(resultName, 'My Ticket Result');
+
+      fireEvent.click(okButton);
+      expect(ticketForm.updateRouter).toBeCalled();
+      expect(ticketForm.updateRouter).toMatchCallSnapshot();
+    });
+  });
+});
```
ericnewcomer, line 50: 👍

PR #885, opened 2020-06-09
🌎 Button to change flow language

rowanseymour:
Captura de Pantalla 2020-06-09 a la(s) 13 40 43
Captura de Pantalla 2020-06-09 a la(s) 13 41 04

### DIFF
--- a/src/test/config.ts
+++ b/src/test/config.ts
@@ -32,7 +32,8 @@ export const config: FlowEditorConfig = {
     activity: '',
     simulateStart: '',
     simulateResume: ''
-  }
+  },
+  onChangeLanguage: (code: string, name: string) => {}
 };
 
 export default config;
--- a/src/components/translator/TranslatorTab.module.scss
+++ b/src/components/translator/TranslatorTab.module.scss
@@ -147,6 +147,11 @@
         color: $light_gray_2;
         border: solid 1px $light_gray_2;
         width: 100%;
+
+        &:disabled, &[disabled] {
+          cursor: default;
+          opacity: 0.4;
+        }
       }
     }
   }
--- a/src/components/translator/TranslatorTab.module.scss
+++ b/src/components/translator/TranslatorTab.module.scss
@@ -66,8 +66,10 @@
   }
 
   &.change_visible {
-    .filters {
-      height: 65px;
+    .changeLanguage {
+      height: 20px;
+      padding: 6px;
+      opacity: 1;
     }
   }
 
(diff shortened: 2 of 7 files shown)


ericnewcomer:
Would much rather there not be a disabled button (not super obvious that button is disabled when not side by side as well). My preference would be no button at all in that case.

rowanseymour:
Hmm I had it that way and it felt like nobody ever is going to discover that feature if that button only appears if you hit 100%... is there anything we can do to make it clearer that the button is disabled?

ericnewcomer:
This seems broken on the netlify preview.

Translation tab is misaligned and button is missing. Maybe from missing configuration?

ericnewcomer:
Disabled buttons are similarly annoying in that they don't necessarily know how to get it enabled in this case. UX people would say keep it enabled but give messaging if they click it when it's not allowed, but I've always hated that paradigm. 

In this case, this feature is effectively being added for u-report who know it exists. I have no problem with it being hidden until it's ready.

Another option is a message telling them what to do to get the option.

src/test/config.ts
```diff
@@ -32,7 +32,8 @@ export const config: FlowEditorConfig = {
     activity: '',
     simulateStart: '',
     simulateResume: ''
-  }
+  },
+  onChangeLanguage: (code: string, name: string) => {}
```
ericnewcomer, line 36: I like this passing of event handlers for configuration, in this case, I guess it would need to trigger a page reload after doing the work at some endpoint?

rowanseymour: yeah that'll throw up a modal to explain what's going to happen, then that POSTs to a django view, calls to mailroom to do the change, then refreshes the editor with new definition

rowanseymour:
I think the netlify issue is me using `opacity: 50%` instead of `opacity: 0.5`.. not sure why that worked locally for me.

rowanseymour:
It's hard to say how many people will use this feature but I don't love assuming that only a certain set of customers could use something. And in this particular case, all of this functionality is already hidden in that you have to click the expand button to see any of it. Then for most users you would have to also uncheck both the categories/rules checkboxes before it appears.

Other option here is to say the button is always enabled _and_ you can switch the flow any time you want, but just warn people that missing translations will keep their original text. Not sure about that.. switching to a 0% translated language is pretty weird... unless the flow had the wrong language to start off with?

Or to your last suggestion, put some text at the bottom that explains something like _"Once fully translated this can be made the default language"_. I think I like that.. and it's still not even going to be visible unless a user hits the expand button.

One trickiness tho is that the expand toggle thing assumes a fixed height for all of this.

ericnewcomer:
Fine with it being always enabled with a warning. The 86% screenshot above has issues though, is that how it renders? ie, the button is pushing the width beyond the scroll window by quite a bit.

rowanseymour:
ahhh I didn't realise my changes made it that way.. I was wondering why it was like that.. will investigate!

rowanseymour:
@ericnewcomer ok I think this is fixed now.. the layout issue wasme using `width: 100%` to try to force a new line in the flexbox.. now I've just moved that into a div below the flex box

Issue #889, opened 2020-07-22
Blocking contacts


rowanseymour:
For https://github.com/rapidpro/rapidpro/issues/607 I've so far added a generic action to the engine to change contact status https://github.com/nyaruka/goflow/issues/939

But not completely sure how best to surface that in the editor.

1. Add to `Update Contact`: you would select `Status` in the same way you select `Language` or `Name`.. and then select one of `Active`, `Blocked` or `Stopped`
2. Explicit UI node types for `Block Contact`, `Stop Contact`.. and maybe  `Activate Contact` ?

Downside of 1 is whether users looking to block a contact would ever think to look for it there. Downside of 2 is adding to the mega list of node options.

nicpottier:
Ya, the mega list of possible actions is mega. I do wonder if we are going to have to rethink that at some point as 1) definitely seems like the right approach here but as you said then decreases discoverability. Maybe being clever about some kind of "searching" in that list could help but that feels cheap.

In any case, I think 1) is right. I hadn't previously thought we would add `Stopped` in there as that is something the contact themselves control but now that I'm thinking of it it does seem to make sense that we allow flow authors to manually stop contacts based on some action the contact is making.

Related, do we think Contact deletion will its own action?

rowanseymour:
I think stop makes sense where it's clear the contact wants to stop communicating and you want to record that and not confuse it with the contact sending you spam and you blocking them. That said, it's a little odd in that the onus is on you the flow designer to record that intention correctly.. so I think the select options need to explain the meaning, e.g. something like..

```
* Blocked (the contact is sending spam etc)
* Stopped (the contact is opting out of communication)
```

I'm still not entirely convinced about making contact deletion part of flows (guess we should chat!) but I don't think it would be part of this node if it were.

rowanseymour:
Re the mega list.. if we said we only allow Block and Stop then I could imagine having one extra item in the mega list like 

```
Block or stop the contact
```

Then on that modal you pick between the two.. @ericnewcomer thoughts?

ericnewcomer:
One quick thought on this is that `Blocked` vs `Stopped` sure sound like the same thing to the naive user. I feel like maybe `Stopped` should be presented to the user as `Opted Out` and they should be able to put people in that status as well. Only reservation I have there is it suggests maybe they are opted in by virtue of not being in that list. I guess there is a difference between an explicit opt out and not being opted in yet though, so maybe thats okay. 

[...]

PR #893, opened 2020-07-29
Block and stop

rowanseymour:
Adds #889 

Captura de Pantalla 2020-07-29 a la(s) 14 03 50


nicpottier:
Thought we'd decided to avoid using "opt-out" because it will collide with possible concepts of "opting-in"?

Occurred to me that the distinction between these two states that maybe is important to users is that a contact that is blocked that sends an incoming message will be ignored while an stopped contact will go through normal flow handling for that message. (and become unstopped in the process)

Not entirely sure how to summarize that though!

rowanseymour:
Oh I thought we'd only decided not to call the status "Opt out" or have "Opt in" as a status.

I think the important thing to explain on this modal, is not all the ways in which blocking and stopping and treated differently in the system, but the fundamental distinction that stopping is the contact opting to end communication, and blocking is you the user opting to end communication.

We might want to link to documentation from the modal if we want users to understand how the system treats these statuses.

nicpottier:
Ya, it may be a lot to explain all at once. What about:

```
blocked - remove from groups, ignore messages
stopped - remove from groups
```

rowanseymour:
As I said I think the more important thing to explain on this modal is the distinction between contact initiated and user initiated. Stopping will happen automatically on some channels if a contact wants to opt-out, but this is a chance in a flow to see that a contact wants to opt-out. 

It's going to get weird if users don't understand this is the same thing that happens if a channel supports stopping/opting-out.

Maybe..

```
Blocked - we want to ignore messages from this contact
Stopped - contact is opting out and will be removed from groups
```

rowanseymour:
This is what we say on the Stopped page:

Captura de Pantalla 2020-07-29 a la(s) 15 38 22

We should probably have a similar blurb on https://app.rapidpro.io/contact/blocked/

nicpottier:
Well I mean it is still going to say "Blocked" and "Stopped" there so I don't think there's any way diminishment in communicating that. I don't think we generally use "we" in the system, so maybe that's what getting my goat. But in this case it really isn't that the user is opting out for example, the flow author is opting them out for whatever reason, so telling them what changing that status does seems more logical to me. And describing the group behavior and msg behavior as I tried seemed like the simplest way to do that.

We are doing a lot more than just ignoring messages for the contact by selecting "Blocked", making it clear that they will be removed from all groups seems like something we should be communicating to users.

@ericnewcomer thoughts here?

rowanseymour:
Well the text I have in the PR is `Stopped - the contact wants to opt-out` which is accurate, if the flow author is using this correctly. I dunno.. I think we want to ensure flow authors use it consistently with how we use it in message handling. Just telling them `Stopped - remove from groups` makes it sound it like a handy way to remove a contact from groups.

nicpottier:
Don't you think removing the contact from all groups is a destructive enough thing that we should be mentioning it front and center? I'm just worried that changing status sounds like a pretty benign thing but really is doing some serious shit that you can't undo.

rowanseymour:
That's a fair point.. tho it leaves me wondering if the block button on the contact list page warrants a health warning.. or we should figure out a way to not remove from groups. 

Maybe we stop trying to stuff all this into the select box and add an actual warning text to the modal, e.g.

Captura de Pantalla 2020-07-29 a la(s) 17 20 10

ericnewcomer:
Maybe something like..
```
blocked - remove from groups, ignore forever
stopped - remove from groups, ignore until they message next
```

With or without the `remove from groups` part depending on how important we feel that is.

rowanseymour:
@ericnewcomer you don't like the idea of pulling that out into an actual warning so we don't have to worry about word count here? 

(maybe the portuguese version will be longer than the select box is wide 🤷)

ericnewcomer:
Personally, I prefer this information to be as concise as possible and next to the decision point. It'll wrap if it needs to with localization.

rowanseymour:
Fine, ok we all happy with..

Captura de Pantalla 2020-07-29 a la(s) 18 10 04

ericnewcomer:
Sure. If we don't want to style it custom, I'd rather we hyphenate instead of using parentheses.

nicpottier:
Either works for me!

rowanseymour:
Updated

Captura de Pantalla 2020-07-30 a la(s) 09 34 52

Issue #628, opened 2019-06-13
Flow type constants don't quite match engine types


rowanseymour:
Was trying to simplify engine vs database types in RP (https://github.com/nyaruka/rapidpro/commit/288e0661bdfccce4c6098ac52881f5360767b6a1) and I noticed the types the editor uses don't quite match the engine types.

Engine:

* messaging
* messaging_offline
* voice

Editor:

* message
* message_offline
* voice

Thankfully it doesn't seem to cause any problems because RP creates the flow with the engine type, and the editor doesn't change it.

ericnewcomer:
Odd, I swear I had changed that exact bit. Maybe I changed it to be wrong!

PR #942, opened 2020-12-16
Use engine flow type names

rowanseymour:
Addresses https://github.com/nyaruka/floweditor/issues/628. Might as well make these match for simplicity before we add another one.

Requires tweak on RP side but it's a nice simplification https://github.com/nyaruka/rapidpro/pull/3126

### DIFF
--- a/src/components/flow/Flow.test.ts
+++ b/src/components/flow/Flow.test.ts
@@ -68,7 +68,7 @@ describe(Flow.name, () => {
       waitNode,
       waitNode.node.exits[0].uuid,
       1,
-      FlowTypes.MESSAGE
+      FlowTypes.MESSAGING
     );
 
     const actionNode = nodes[nodeMapKeys[0]];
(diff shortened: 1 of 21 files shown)


rowanseymour:
I think I need to tweak https://github.com/nyaruka/floweditor/blob/7fd36e70521e9166cd0fe48ffc86c064b9c0dbb7/src/components/flow/routers/subflow/SubflowRouterForm.tsx#L109

That's currently not quite working - if you're in a Surveyor flow it seems you can enter any other kind of flow.

ericnewcomer:
Argh, I think you've discovered another case where the big asset selector change introduced a bug. That filter should be working on native objects as they are returned from the endpoint, not the converted asset. I'll update this PR for that.

ericnewcomer:
Oops, I take that back! This form is still using the old AssetSelection stuff, so not a bug (but would be nice to have it use temba-select based stuff like everything else (and we could likely remove react-select as a dependency in the process).

So, to answer your question, as long at the flows endpoint is returning the same things for flow_type nothing should need to change there. I pulled the corresponding rapid PR and can't reproduce the issue of surveyor flows being able to enter non-surveyor flows.

ericnewcomer (approved):

rowanseymour:
I see the problem of not filtering on prod, but hadn't yet got editor working in dev mode locally to test with these changes

PR #944, opened 2020-12-17
Fix flow filter

ericnewcomer:


src/external/index.ts
```diff
@@ -173,6 +173,17 @@ export const getAssets = async (url: string, type: AssetType, id: string): Promi
   return assets;
 };
 
+export const getFlowType = (flow: any) => {
```
rowanseymour, line 176: seems like this could replace the switch below in `resultToAsset` so it's clear that isn't used any more

ericnewcomer: Ya, resultToAsset is going away. Didn't have time to do an audit.

Issue #960, opened 2021-01-28
Add support for dial node


rowanseymour:
Not sure tho what the most user-friendly term for this would be.. "Forward Call", "Redirect Call", "Dial Number"

PR #961, opened 2021-01-28
☎️ Implement dial router

rowanseymour:
Adds https://github.com/nyaruka/floweditor/issues/960

Captura de Pantalla 2021-02-01 a la(s) 09 34 39

Captura de Pantalla 2021-02-01 a la(s) 09 34 02

Captura de Pantalla 2021-02-01 a la(s) 09 34 57


src/config/typeConfigs.ts
```diff
@@ -210,6 +211,14 @@ export const typeConfigList: Type[] = [
     visibility: [FlowTypes.MESSAGING_OFFLINE, FlowTypes.VOICE]
   },
 
+  {
+    type: Types.wait_for_dial,
+    name: i18n.t('actions.wait_for_dial.name', 'Wait for Call'),
+    description: i18n.t('actions.wait_for_dial.description', 'Wait for a call to another number'),
+    form: DialRouterForm,
+    visibility: VISIBILITY_VOICE
```
rowanseymour, line 219: @ericnewcomer was thinking to have a component which renders the phone number.. but it seems like components are setup for actions and the other splits don't render anything, so not sure..

src/components/flow/routers/dial/DialRouterForm.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        $merge: { updateRouter: jest.fn(), onClose: jest.fn() }
+      });
+
+      instance.handlePhoneUpdated('+1234567890');
+      instance.getButtons().secondary.onClick();
+      expect(props.onClose).toHaveBeenCalled();
+      expect(props.updateRouter).not.toHaveBeenCalled();
+    });
+  });
+});
```
ericnewcomer, line 56: This is modeled after legacy tests. Instead new code should use `react-testing-library` which doesn't let you interact with the instance instead forcing you to use user events to test. 

Here's an example here:
https://github.com/nyaruka/floweditor/blob/main/src/components/flow/routers/result/ResultRouterForm.test.tsx

src/config/typeConfigs.ts
ericnewcomer: As in a temba-component? This seems like a useful thing to include over there (which we could then use directly here) `<temba-phone/>`

ericnewcomer (approved):
Would like the test to use react-testing-library, but that can also come later as we have a lot of tests still using the current `composeCompontentTests...` approach. I should add a deprecation notice to avoid this confusion.

Are we ready to ship this? Shall I merge it in for today's release?

src/config/typeConfigs.ts
rowanseymour: No the component which renders content on the node (e.g. `SendMsgComp`). But we don't do this for other split types (e.g. you can't see the expression on a Split By Expression node) so I think it's ok without it.

rowanseymour:
@ericnewcomer we can't merge this yet because we don't have a way to hide it and @nicpottier is still working on the mailroom side

src/config/typeConfigs.ts
ericnewcomer: Ah, well, in any case, seems useful enough to have it live in temba-components so we could use it everywhere.

ericnewcomer:
I suppose we could with an empty visibility flag, but no reason to do it just yet either way

ericnewcomer (approved):

PR #968, opened 2021-03-02
Fix editor language detection and add Spanish (es) translation

rowanseymour:


src/config/i18n/index.js
```diff
@@ -8,30 +8,29 @@ import LanguageDetector from 'i18next-browser-languagedetector';
...
   .use(LanguageDetector)
   .use(initReactI18next)
   .init({
-    resources,
-
-    // do not load a fallback
-    fallbackLng: ['en', 'defaults'],
-
+    detection: {
+      order: ['querystring', 'htmlTag']
```
rowanseymour, line 20: Noticed default settings we're letting browser override the RP setting. Query string just here for testing so you can add `lng` to the URL to try a different language

src/config/i18n/index.js
```diff
@@ -8,30 +8,29 @@ import LanguageDetector from 'i18next-browser-languagedetector';
...
+      fr: { translation: fr },
+      'pt-br': { translation: ptBR }
+    },
+    fallbackLng: ['defaults'],
     interpolation: {
       prefix: '[[',
       suffix: ']]'
-    }
+    },
+    lowerCaseLng: true
```
rowanseymour, line 33: django language code set on `<html>` tag is lowercase like `pt-br` - this is needed to make detection work properly

ericnewcomer (approved):

PR #977, opened 2021-03-23
Fix: wrong attribute comparison for templates variables on change

paulobernardoaf:
The attribute `id` does not exist in neither `template` and `this.state.template.value`. 
Both of them have the `uuid` attribute.
The comparison was `undefined !== undefined`, leading to always return the `this.state.templateVariables` for `templateVariables`.

Closes: https://github.com/rapidpro/rapidpro/issues/1432


ericnewcomer:
We don't actually take PRs on this particular repo, but we'll get this fixed up shortly, thanks!

PR #1002, opened 2021-06-30
Adjust treatment for open ticket action for internal ticketers

ericnewcomer:


src/components/flow/actions/openticket/OpenTicket.tsx
```diff
@@ -1,17 +1,21 @@
...
-      id: ticketer.uuid,
-      name: ticketer.name,
-      type: AssetType.Ticketer
-    },
-    context.config.endpoints
+const OpenTicketComp: React.SFC<OpenTicket> = (
+  { ticketer, subject },
+  context: any
+): JSX.Element => {
+  const showTicketer = ticketer.name.indexOf(context.config.brand) === -1;
```
rowanseymour, line 9: maybe more reliable to check the ticketer type? Just thinking that Zendesk and Mailgun ticketer names could actually include strings like rapidpro and textit.. if that's what people call their accounts on those services.

ericnewcomer: Yes. Meant to add a comment here, this is super goofy.

The definition doesn't include the ticketer type. Looked at adding it to the ui metadata, but the editor deals with generic asset selection with name/uuid here and would be a big refactor to change that.

ericnewcomer: Not to mention the rendering of tickets only has the context of the current action. This switch is simply for whether we show the subtext of which ticketer it is opening a ticket against. So I think it is okay if this is hidden in that edge case for now.

ericnewcomer: Alternatively, I considered just always showing the ticketer used, but that feels worse for the common case.

rowanseymour: I thought the plan was to hide/show it based on just whether there are > 1 ticketers ?

ericnewcomer: You are confusing this with the form. This is the rendering of the flow. The form select shows if there are > 1 ticketer or if the currently selected ticketer no longer exists.

src/components/flow/routers/ticket/TicketRouterForm.tsx
```diff
@@ -136,20 +136,31 @@ export default class TicketRouterForm extends React.Component<
   private renderEdit(): JSX.Element {
     const typeConfig = this.props.typeConfig;
 
+    // if we only have one ticketer or we have issues, show the ticket chooser
+    const showTicketers =
+      Object.keys(this.props.assetStore.ticketers.items).length > 1 || this.props.issues.length > 0;
```
ericnewcomer, line 141: This is the switch you are thinking of..

src/components/flow/actions/openticket/OpenTicket.tsx
rowanseymour: aaaahhhh in which case this is goofy but completely fine

Issue #1014, opened 2021-09-27
Hide location functions when `locations` feature flag not set


rowanseymour:
`has_state`, `has_district`, `has_ward`

PR #1016, opened 2021-10-05
Only show location router tests if locations feature flag set

rowanseymour:
Adds https://github.com/nyaruka/floweditor/issues/1014


ericnewcomer (approved):
Bummer that netlify seems broken again, but looks good

PR #1024, opened 2021-11-08
Use child.status rather than child.run.status

rowanseymour:
In the engine `child.run.*` was only kept around for backwards compatibility and isn't exposed in autocomplete.


src/components/simulator/ContextExplorer.tsx
```diff
@@ -131,10 +125,6 @@ export default class ContextExplorer extends React.Component<
     }
 
     const newPath = [...path, name];
-    if (EXCLUDED_PATHS[newPath.join('.')]) {
```
rowanseymour, line 134: this isn't needed anymore - `child.run` and `parent.run` aren't in the autocomplete files

PR #1027, opened 2022-01-03
Simulator should use wait events rather than the wait object

rowanseymour:
Trying to get rid of wait objects on sessions because it's just duplication of what is in the wait event. Mailroom doesn't use them.

### DIFF
--- a/src/components/simulator/LogEvent.tsx
+++ b/src/components/simulator/LogEvent.tsx
@@ -3,7 +3,7 @@ import { MediaPlayer } from 'components/mediaplayer/MediaPlayer';
 import Modal from 'components/modal/Modal';
 import styles from 'components/simulator/LogEvent.module.scss';
 import { Types } from 'config/interfaces';
-import { Flow, Group, Label, Topic } from 'flowTypes';
+import { Flow, Group, Label, Topic, Hint } from 'flowTypes';
 import * as React from 'react';
 import { createUUID, getURNPath } from 'utils';
 import i18n from 'config/i18n';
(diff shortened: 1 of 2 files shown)


src/components/simulator/LogEvent.tsx
```diff
@@ -72,6 +72,8 @@ export interface EventProps {
   classifier?: { uuid: string; name: string };
   ticketer?: { uuid: string; name: string };
   ticket?: { topic: Topic; body: string };
+  hint?: Hint;
+  timeout_seconds?: number;
```
rowanseymour, line 76: it'd be pretty neat if the simulator simulated timeouts and you could see a timer counting down to when you get moved on

src/components/simulator/Simulator.tsx
```diff
@@ -131,7 +131,6 @@ interface Run {
   flow_uuid: string;
   status: string;
   events?: EventProps[];
-  wait?: Wait;
```
rowanseymour, line 134: runs have never had waits

src/components/simulator/Simulator.tsx
```diff
@@ -145,8 +144,7 @@ interface Session {
   runs: Run[];
   contact: Contact;
   input?: any;
-  wait?: Wait;
-  status?: string;
```
rowanseymour, line 149: status is always set on sessions

src/components/simulator/LogEvent.tsx
ericnewcomer: We talked about this in the past, but giving the user the option to trigger no-response. Could do both.

ericnewcomer (approved):

PR #1073, opened 2022-09-19
Revert netlify change

ericnewcomer:


ericnewcomer:
@liuzl realized your change was actually incorrect, but this PR is to correct it and make sure netlify is working.

Issue #1081, opened 2022-10-24
Translating "has number between" rules only shows one input


rowanseymour:
But there are 2 args. Also editor currently saves a translation which is `[""]` if you just press save without entering values.

rowanseymour:
Actually we should just disable translations for all numerical rules...

susanm74:
@rowanseymour for this issue, should i go ahead and fix the 2 bugs described in your first comment? OR should i hold off until we can roughly sketch out what a "numerical translations disabled" ux would look like? for the latter, i'm thinking:
- when a user toggles a different language, text for untranslated numerical rules remains black instead of red
- when a user toggles a different language, and clicks a node, instead of displaying the router localization form, nothing happens
- would "has a date..." and "has a time..." also be considered numerical and thus also not necessary to translate?

susanm74:
@rowanseymour ok fixed the has number between issue, unit test in progress, and then will go ahead and work on the saving empty translations issue and bundle that in with the same PR, unless it's preferred to make a separate PR for that? let me know!)

rowanseymour:
We definitely want to fix saving empty translations so go ahead there, but re the the has_number_between issue, I'll let @ericnewcomer confirm but I think that just shouldn't appear as an option to translate.

susanm74:
@rowanseymour i'm not sure that i'm able to reproduce the empty translations issue - can you let me know if there's something i'm missing or doing wrong?... when i step through, i'm not seeing any blank or empty message text being saved as part of the localizations object - see the 3 screenshots below in which i've set a breakpoint in the handleSave fn to check what the localizations object contains right before it's passed to the updateLocalizations fn -
1. when i click the node to translate, do nothing, and click ok:
<img width="678" alt="Screen Shot 2022-10-26 at 1 08 01 PM" src="https://user-images.githubusercontent.com/854012/197948405-b8699766-8916-49f7-a98e-b102ca4e539f.png">

2. when i click the node to translate, type in 'blah', and click ok:
<img width="547" alt="Screen Shot 2022-10-26 at 1 08 49 PM" src="https://user-images.githubusercontent.com/854012/197948528-8da286d5-bef4-4dbd-bf25-295a2762d0d6.png">

3. when i click the node to translate, delete 'blah', and click ok:
[...]

PR #1082, opened 2022-10-25
wait for response form - numeric rule - rule translations tab - has number between and empty translations bugfixes

susanm74:
https://github.com/nyaruka/floweditor/issues/1081


src/components/flow/routers/localization/RouterLocalizationForm.tsx
```diff
@@ -119,7 +119,9 @@ export default class RouterLocalizationForm extends React.Component<
       const { verboseName } = getOperatorConfig(originalCase.type);
 
       if (originalCase.arguments) {
-        const [orginalArgument] = originalCase.arguments;
```
susanm74, line 122: i tried googling, but couldn't find any refs or examples of declaring a variable this way `const [orginalArgument]` - is this syntax some sort of var declaration shorthand?

susanm74:
has number between bugfix
BEFORE
Screen Shot 2022-10-27 at 1 18 09 PM
AFTER
Screen Shot 2022-10-27 at 1 22 39 PM

susanm74:
saving empty rule translations bugfix
BEFORE
Screen Shot 2022-10-27 at 1 20 22 PM

AFTER
Screen Shot 2022-10-27 at 1 24 14 PM

susanm74:
@ericnewcomer @norkans7 @rowanseymour a gentle nudge here

ericnewcomer:
What do we do in the rule localization? I thought it was the argument list that was the primary issue.

rowanseymour:
I thought we said we were going to get rid of localizing numerical rule arguments...

ericnewcomer:
Ya, my understanding is that the fix we really need is the filtering out of has_number_between rules from the list of localizable arguments.

susanm74:
AFTER - with all "has number between" categories - rule translations tab hidden
Screen Shot 2022-11-02 at 2 42 59 PM
Screen Shot 2022-11-02 at 2 39 12 PM

AFTER - with a mix of "has number between" categories and other categories - rule translations tab displays but filters those out from the list
Screen Shot 2022-11-02 at 2 43 10 PM
Screen Shot 2022-11-02 at 2 39 23 PM
Screen Shot 2022-11-02 at 2 39 36 PM

susanm74:
@ericnewcomer @norkans7 @rowanseymour RFR again, now also includes the fix for filtering out "has number between" rules from the rule translations list

norkans7:
What about the other numerical rules types such as `Operators.has_number_lt`, `Operators.has_number_gt`?
I think their arguments as well should not be translated

susanm74:
@norkans7 i think we're just targeting this numeric rule for now because it's the only numeric rule that involves 2 arguments (ranges with an lower and upper bound), all of the others only contain 1 argument, and thus, aren't causing the issues this 2 argument numeric rule was causing, but yeah, i do think if we wanted to exclude those as well, we could easily exclude those from translation in a follow-up PR

src/components/flow/routers/localization/RouterLocalizationForm.tsx
```diff
@@ -194,7 +198,11 @@ export default class RouterLocalizationForm extends React.Component<
 
     const hasCasesWithArguments = !!this.state.cases.find((kase: Case) => {
       const orginalCase = getOriginalCase(this.props.nodeSettings, kase.uuid) as Case;
-      return orginalCase.arguments && orginalCase.arguments.length > 0;
+      return (
+        orginalCase.arguments &&
+        orginalCase.arguments.length > 0 &&
+        orginalCase.type !== Operators.has_number_between
```
ericnewcomer, line 204: This one-liner would be more robust:
`(originalCase.arguments || []).length == 1`

PR #1083, opened 2022-10-27
start somebody else in a flow form - add skip contacts in a flow checkbox

susanm74:
BEFORE
Screen Shot 2022-10-27 at 2 34 45 PM

### DIFF
--- a/src/components/flow/actions/startsession/helpers.ts
+++ b/src/components/flow/actions/startsession/helpers.ts
@@ -34,6 +34,7 @@ export const initializeForm = (settings: NodeEditorSettings): StartSessionFormSt
           : START_TYPE_ASSETS
       },
       contactQuery: { value: action.contact_query || '' },
+      skipContactsInFlow: false,
       valid: true
     };
 
@@ -45,6 +46,7 @@ export const initializeForm = (settings: NodeEditorSettings): StartSessionFormSt
     flow: { value: null },
     startType: { value: START_TYPE_ASSETS },
     contactQuery: { value: '' },
+    skipContactsInFlow: false,
     valid: false
   };
 };
--- a/src/components/flow/actions/startsession/StartSessionForm.module.scss
+++ b/src/components/flow/actions/startsession/StartSessionForm.module.scss
@@ -7,3 +7,7 @@
   margin-top: 0.5em;
   margin-bottom: 1em;
 }
+
+.checkbox {
+  margin-top: 20px;
+}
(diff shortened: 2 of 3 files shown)


susanm74:
AFTER

susanm74:
@rowanseymour @norkans7 @ericnewcomer if a user checks the "skip contacts currently in a flow" checkbox as part of the start somebody else in a flow action, is there another repo where i need to added a check to see if this checkbox was checked before contacts are started in a flow? i was poking around in the rapidpro FlowCRUDL Broadcast class, but wasn't sure whether there or in a different repo or both?

norkans7:
goflow already has support for that but for that to work you need to follow https://github.com/rapidpro/rapidpro/issues/1662#issuecomment-1289285797

So instead of using `skipContactsInFlow: true` you need to use  `exclusions: {"in_a_flow": true}`

And the action editor should use the advanced tab similar to them one shown below

Monosnap test 2 2022-10-27 12-44-03

Monosnap test 2 2022-10-27 12-44-31

susanm74:
@norkans7 ah ok cool thanks for the review of what’s there so far and for the clarification on the UX and logic!

src/components/flow/actions/startsession/helpers.ts
```diff
@@ -62,7 +64,8 @@ export const stateToAction = (
     create_contact: state.startType.value === START_TYPE_CREATE,
     flow: { name: flow.name, uuid: flow.uuid },
     type: Types.start_session,
-    uuid: getActionUUID(settings, Types.start_session)
+    uuid: getActionUUID(settings, Types.start_session),
+    skipContactsInFlow: state.skipContactsInFlow
```
ericnewcomer, line 68: this is the wrong format

src/flowTypes.ts
```diff
@@ -455,6 +455,7 @@ export interface StartSession extends RecipientsAction {
   flow: Flow;
   create_contact?: boolean;
   contact_query?: string;
+  skipContactsInFlow: boolean;
```
ericnewcomer, line 458: same here

src/testUtils/assetCreators.ts
```diff
@@ -181,19 +181,22 @@ export const createStartSessionAction = ({
...
-  create_contact = false
+  create_contact = false,
+  skipContactsInFlow = false
 }: {
   uuid?: string;
   groups?: Group[];
   contacts?: Contact[];
   flow?: Flow;
   create_contact?: boolean;
+  skipContactsInFlow?: boolean;
```
ericnewcomer, line 192: and here

susanm74:
AFTER
Screen Shot 2022-11-03 at 2 49 36 PM
Screen Shot 2022-11-03 at 2 49 45 PM

PR #1084, opened 2022-10-28
Allow wait for digits routers to use `has_text` and `has_pattern` rules

rowanseymour:


src/config/interfaces.ts
```diff
@@ -85,15 +85,15 @@ export enum Operators {
...
   has_district = 'has_district',
   has_state = 'has_state',
   has_ward = 'has_ward',
   has_pattern = 'has_pattern',
   has_error = 'has_error',
   has_intent = 'has_intent',
   has_top_intent = 'has_top_intent',
-  has_category = 'has_category'
+  has_category = 'has_category',
+  has_value = 'has_value' // deprecated, use has_text
```
rowanseymour, line 96: assuming I can't just remove this because there are flows out there using it

ericnewcomer: Unless we add a flow version migration. Remember those?

ericnewcomer (approved):

Issue #1034, opened 2022-01-26
Shouldn't show Expired exit for IVR enter flow actions


rowanseymour:
When IVR flows expire, calls are hung up and we don't resume parent flows.

<img width="230" alt="Captura de Pantalla 2022-01-26 a la(s) 13 52 24" src="https://user-images.githubusercontent.com/675558/151253168-87bae3c6-f16d-4de6-9da5-5374083de001.png">

susanm74:
ingresar en un flujo => enter a flow

rowanseymour:
@rowanseymour sorry what I meant in the original comment is that _we currently hang up calls and don't resume parent flows_, thus the Expired exit is never used in an IVR flow. Thus the only thing needed here is not showing it in the editor for IVR flows.

ericnewcomer:
Fixed in #1086

PR #1086, opened 2022-11-11
remove expired exit from ivr enter flow action

susanm74:
#1034


susanm74:
BEFORE

norkans7 (approved):

norkans7 (changes requested):
what makes the `expired` category appear for messaging flows?

susanm74:
> what makes the `expired` category appear for messaging flows?

@norkans7 ah i see, will take a look at making sure the expired exit doesn't disappear for msg enter flow actions

susanm74:
AFTER - msg flow type
Screen Shot 2022-11-21 at 3 42 01 PM

AFTER - voice flow type
Screen Shot 2022-11-21 at 3 41 32 PM

src/components/flow/actions/startsession/__snapshots__/StartSessionForm.test.tsx.snap
```diff
@@ -935,6 +937,7 @@ Array [
     },
     "flow": Object {
       "name": "My Flow",
+      "type": "flow",
```
susanm74, line 940: this feels like a hint that something with my testing approach is off, as here i think the type should either be "messaging" or "voice" and not "flow

src/components/flow/actions/startsession/__snapshots__/StartSessionForm.test.tsx.snap
```diff
@@ -990,6 +993,7 @@ Array [
...
```
susanm74, line 996: this also feels like a hint that something with my testing approach is off, as here i think the type should either be "messaging" or "voice" and not "flow

src/components/flow/routers/subflow/__snapshots__/SubflowRouterForm.test.tsx.snap
```diff
@@ -21,6 +21,7 @@ Array [
         Object {
           "flow": Object {
             "name": "My Subflow Flow",
+            "type": undefined,
```
susanm74, line 24: this also feels like a hint that something with my testing approach is off, as here i think the type should either be "messaging" or "voice" and not "undefined"

susanm74:
@ericnewcomer @norkans7 @rowanseymour ok so the core logic is finished and working and i've manually tested, but i think all of the existing unit tests assume a messaging flow? but for whatever reason it's been trickier than i thought to integrate a voice flow - would love your feedback on whether my testing approach is on the right track or whether i should switch gears and take a different approach with those, bc i can sense something is off, i'm just not quite sure what

src/components/flow/actions/startsession/__snapshots__/StartSessionForm.test.tsx.snap
susanm74: unless this might be ok as it looks like this is just a flowContext.ts -> AssetType.Flow enum -> "flow"

src/components/flow/actions/startsession/__snapshots__/StartSessionForm.test.tsx.snap
susanm74: unless this might be ok as it looks like this is just a flowContext.ts -> AssetType.Flow enum -> "flow"

src/components/flow/routers/subflow/__snapshots__/SubflowRouterForm.test.tsx.snap
susanm74: actually, i think this one should also be flowContext.ts -> AssetType.Flow enum -> "flow" (and not "undefined") per https://github.com/nyaruka/floweditor/blob/main/src/components/flow/routers/subflow/SubflowRouterForm.test.tsx#L46, so i'm not sure why that's undefined bc i don't think anything changed with that specific mocked object

src/components/flow/actions/startsession/__snapshots__/StartSessionForm.test.tsx.snap
ericnewcomer: If memory serves this is for the generic asset selection stuff. It can work against flows, contacts, groups, etc.

src/components/flow/routers/subflow/SubflowRouterForm.test.ts
```diff
@@ -66,7 +72,76 @@ describe(SubflowRouterForm.name, () => {
...
+    });
+
+    it('should cancel changes', () => {
+      const { instance, props } = voiceSetup(true, { updateRouter: setMock() });
+      instance.handleFlowChanged([SoundFlowAsset]);
+      instance.getButtons().secondary.onClick();
+      expect(props.updateRouter).not.toBeCalled();
+    });
+
+    it('converts from other node types', () => {
```
susanm74, line 114: this test isn't passing but i'm not sure why

src/components/flow/routers/subflow/SubflowRouterForm.test.ts
```diff
@@ -66,7 +72,76 @@ describe(SubflowRouterForm.name, () => {
...
+          originalNode: { ui: { $merge: { type: null } } }
+        }
+      });
+
+      instance.handleFlowChanged([SoundFlowAsset]);
+      instance.getButtons().primary.onClick();
+      expect(props.updateRouter).toMatchCallSnapshot();
+    });
+
+    it('creates its own action uuid if necessary', () => {
```
susanm74, line 127: this test isn't passing but i'm not sure why

src/components/flow/routers/subflow/SubflowRouterForm.tsx
```diff
@@ -89,7 +89,9 @@ export default class SubflowRouterForm extends React.PureComponent<
     );
 
     if (this.state.valid && !hasFieldErrors) {
-      this.props.updateRouter(stateToNode(this.props.nodeSettings, this.state));
+      this.props.updateRouter(
+        stateToNode(this.props.nodeSettings, this.state, this.context.config.flowType)
```
susanm74, line 93: during handleSave, when i tried using `this.context.config.definition.type` i was getting an error that definition is undefined, but when i set a breakpoint, i noticed that `this.context.config.flowType` contains what i believe is the parent flow type - it's a bit unexpected as i thought the idea was to get the flow type from the definition, so here's some more context...

during handleSave, `this.context.config.flow` = `e19aa222-2ca8-4328-a16a-2a9f63911b9f` which matches the flow uuid value in the URL = `e19aa222-2ca8-4328-a16a-2a9f63911b9f`, which i assume is the parent flow's uuid, whereas `this.state.flow.value.uuid` = `cacddc3e-9393-4303-af6c-447b089bb494` which i'm assuming is the subflow's uuid...

and so `this.context.config.flowType` will get the job done, but i just wanted to do a final sanity check to make sure this is the correct arg to pass into stateToNode

susanm74:
@ericnewcomer if you could take another pass at this... everything's done expect for that i couldn't quite figure out why 2 tests are are outputting "expired" exits when they shouldn't be, the snapshot is correct, but when running yarn test, what's received is different from what the snapshot expects, i'm just not sure why...

susanm74:
@ericnewcomer when you get a chance this is RFR again!

src/components/flow/routers/subflow/SubflowRouterForm.test.ts
```diff
@@ -1,56 +1,65 @@
...
       expect(wrapper).toMatchSnapshot();
     });
   });
 
-  describe('updates', () => {
+  describe('msg updates', () => {
     it('should update and save', () => {
-      const { instance, props } = setup(true, { updateRouter: setMock() });
+      const { instance, props } = msgSetup(true, { updateRouter: setMock() });
+      expect(instance.context.config.flowType).toEqual(FlowTypes.MESSAGING);
```
susanm74, line 32: @ericnewcomer i added a line to each test to explicitly check the flowType, but when i uncomment the ivr tests, the msg ones suddenly fail, i'm not sure why, but when i step through, it almost seems as though something like this is happening:
- msg router node and setup are created
- ivr router node and setup are created (but it almost seems like this might be overwriting the msg stuff?)
- msg tests are run (everything passes except for the flow type assertions)
- ivr tests are run (everything passes)
any ideas as to how i can resolve this? i'm hoping maybe it's just something quirky with my test setup?

ericnewcomer:
@susanm74 where are we on this?

susanm74:
@ericnewcomer ooh, sorry for the delay here, will def take a look at this today and confirm if everything's ready to go - i'm 99% sure it is, just want to do a final check to confirm

susanm74:
@ericnewcomer yes, this is RFR and/or merge! 😄

Issue #1054, opened 2022-03-16
Creating a new node after a horizontal scroll causes the new node to move to some distance


mdshamoon:
## Issue
Creating a new node after a horizontal scroll causes the new node to move to some distance

Sharing the recording for reference

https://user-images.githubusercontent.com/32592458/158587261-e0ae5e0c-76b7-4646-89d7-f5ff0dbd51c5.mp4

ericnewcomer:
I'm not seeing this with the current version of the editor. Are you able to reproduce this on TextIt?

mdshamoon:
It's not replicable on TextIt. I am able to replicate by running the editor on my local machine. 
Maybe it gets fixed when used inside TextIt. Some overflow issues I think while just using the editor. 
Thanks for checking.

PR #1173, opened 2023-02-16
Fix: Creating a new node after a horizontal scroll causes the new node to move to some distance

mdshamoon:
## Summary

While trying to fix the https://github.com/nyaruka/rapidpro/issues/4303 this issue https://github.com/nyaruka/floweditor/issues/1054 comes up in editor. 

The issue is replicable when we try to just load the editor without Rapidpro. Just scroll horizontally to some distance and try to add a new node. Have tried this on https://floweditor.nyaruka.com and this issue is present there.

I have added a fix for this by changing the pageX attribute to screenX which seems to fix the issue in this case and helps in fixing the scroll issue when we open revision history. Please let me know if this change looks fine.

### DIFF
--- a/src/components/flow/node/Node.tsx
+++ b/src/components/flow/node/Node.tsx
@@ -129,7 +129,7 @@ export class NodeComp extends React.PureComponent<NodeProps> {
 
         // move our ghost node into position
         const width = this.ele.getBoundingClientRect().width;
-        const left = e.pageX - width / 2 - 15 - canvasBounds.left;
+        const left = e.screenX - width / 2 - 15 - canvasBounds.left;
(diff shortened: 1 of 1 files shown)


mdshamoon:
I have read the CLA Document and I hereby sign the CLA

mdshamoon:
@ericnewcomer please let me know your thoughts on this.

ericnewcomer:
Thanks for this, I'll take a look. Sorry, I was out sick last week.

ericnewcomer:
I did some testing and this doesn't work correctly when embedded. I opened a new PR that I think works both embedded and standalone that I think should do the trick here: https://github.com/nyaruka/floweditor/pull/1174

PR #1183, opened 2023-04-12
Use compose component for Send To Somebody Else action

susanm74:
fixes https://github.com/nyaruka/floweditor/issues/1181
needs to be merged together with https://github.com/nyaruka/temba-components/pull/315


susanm74:
BEFORE
Screen Shot 2023-04-13 at 3 06 35 PM

susanm74:
@ericnewcomer just a quick update:
- if you get a chance and want to see how things are looking in the code so far (note that things are not yet working e2e), some early feedback would be nice reassurance that things are generally headed in the right direction
- would love to chat through some questions i've got about the mechanics of event handling between the floweditor's compose element and the temba-components' compose component (ex. SendBroadcastForm.handleComposeChanged, ComposeElement.ComposeProps.onChange, ComposeElement.handleChange, etc) when i'm back next week

susanm74:
AFTER
Screen Shot 2023-04-24 at 2 32 07 PM

ericnewcomer:
Really close, a few things:

 1) When hovering over the char counter, the details appear far away.. this might be the trickiest thing on this list, so maybe save it for last. If it's not a quick fix, I'm willing to ship this bug instead of wasting a day on it since I suspect it is related to how the dialog is positioned and we really want to replace the editor's react dialogs with `temba-dialog` at some point.
 2) We should limit the number of attachments to three instead of ten. Another thing I noticed is that the react dialogs don't inner scroll, so users can get themselves into troubles selecting a bunch of files to attach. Nothing we can do about that without fixing the dialog I don't think. But just noting it as something I ran into.
 3) The paperclip treatment on the main flow page should match Send Message attachment which has a tilt
 4) Simulating this action should behave like the Send Email action (it's currently rendering like a message, but missing the attachment). It should instead have a link "Sent somebody else a message" and clicking it should show a dialog with the message and any attachments.

susanm74:
@ericnewcomer 
1. ooh ok, i see this, will see what i can do
2. 👍 
3. 👍 
4. ah ok, working on this one

susanm74:
@ericnewcomer this is RFR again... (note - needs to be reviewed/merged with https://github.com/nyaruka/temba-components/pull/315)

ericnewcomer (approved):

PR #1191, opened 2023-07-05
Fix has a ward

norkans7:


ericnewcomer:
This worked well when I tested it. Any tips on how to make it fail?

ericnewcomer (approved):

norkans7:
The issue was that when using has a ward rule, the next time one open the
node editor the stat and district parameters were not being loaded and the
validation was now failing even tho the flow definition was correct. That's
how I understood that

ericnewcomer:
Right, I was able to repro that, but from what I can tell your changes in this PR address it as I can no longer. You seemed unsure in previous comments, so I was just asking before merging/shipping.

norkans7:
I think I was confused because the flow I tried to test with was not saving
but later I was able to confirm that was the correct fix

PR #1256, opened 2024-12-02
Make char limits on result name and category on `set_run_result` actions consistent with routers

rowanseymour:


src/components/flow/actions/setrunresult/SetRunResultForm.tsx
```diff
@@ -112,6 +112,9 @@ export default class SetRunResultForm extends React.PureComponent<
   }
 
   private handleCreateAssetFromInput(input: string): Asset {
+    // workaround for the lack of a length limit on the form itself
+    input = input.substring(0, 64);
```
rowanseymour, line 116: this is the Result Name entered as an arbitrary option on a tembaselect

Issue #1273, opened 2025-04-09
Make result optional on webhook splits


rowanseymour:
By changing the router to not use the result. 

We currently construct webhook routers like:

```json
{
        "type": "switch",
        "operand": "@results.cat_fact.category",
        "cases": [
            {
                "uuid": "190c8a9d-2a22-497a-9bc8-3f11c4af56c3",
                "type": "has_text_eq",
                "arguments": [
                  "Success"
                ],
                "category_uuid": "a8ddc734-6b34-444b-b30b-9838615a095e"
            },
            {
                "uuid": "ff810df6-23c2-4dff-9be1-eebffae2bb9d",
                "type": "has_text_eq",
                "arguments": [
                  "Failure"
                ],
                "category_uuid": "a8ddc734-6b34-444b-b30b-9838615a095e"
            }
        ],
        "categories": [
            {
                "uuid": "a8ddc734-6b34-444b-b30b-9838615a095e",
                "name": "Success",
                "exit_uuid": "d7488301-2b2a-4562-9d22-3dfb7bef0600"
            },
            {
                "uuid": "5f2c6bb7-0473-4ff1-abf0-5199c0ff87a9"
                "name": "Failure",
                "exit_uuid": "251a9293-85c2-473a-b302-c0f79624d9bb"
            }
        ],
        "default_category_uuid": "5f2c6bb7-0473-4ff1-abf0-5199c0ff87a9"
    }
```

We would construct them like this instead:

```json
{
        "type": "switch",
        "operand": "@webhook.status",
        "cases": [
            {
                "uuid": "190c8a9d-2a22-497a-9bc8-3f11c4af56c3",
                "type": "has_number_between",
                "arguments": [
                  "200",
                  "299"
                ],
                "category_uuid": "a8ddc734-6b34-444b-b30b-9838615a095e"
            }
        ],
        "categories": [
            {
                "uuid": "a8ddc734-6b34-444b-b30b-9838615a095e",
                "name": "Success",
                "exit_uuid": "d7488301-2b2a-4562-9d22-3dfb7bef0600"
            },
            {
                "uuid": "5f2c6bb7-0473-4ff1-abf0-5199c0ff87a9"
                "name": "Failure",
                "exit_uuid": "251a9293-85c2-473a-b302-c0f79624d9bb"
            }
        ],
        "default_category_uuid": "5f2c6bb7-0473-4ff1-abf0-5199c0ff87a9"
    }
```

[...]

PR #1274, opened 2025-04-10
Rework webhook nodes to route on @webhook.status

rowanseymour:
Part 1 of https://github.com/nyaruka/floweditor/issues/1273

### DIFF
--- a/src/components/flow/routers/airtime/__snapshots__/AirtimeRouterForm.test.ts.snap
+++ b/src/components/flow/routers/airtime/__snapshots__/AirtimeRouterForm.test.ts.snap
@@ -156,7 +156,7 @@ Array [
             ],
             "category_uuid": "763e4844-3e1b-407a-a1b5-5fdfcd308b41",
             "type": "has_category",
-            "uuid": "1e47a1e1-3c67-4df5-adf1-da542c789adb",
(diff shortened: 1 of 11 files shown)


src/flowTypes.ts
```diff
@@ -585,16 +585,11 @@ export enum StartFlowExitNames {
   Expired = 'Expired'
 }
 
-export enum WebhookExitNames {
+export enum ServiceCallExitNames {
   Success = 'Success',
   Failure = 'Failure'
 }
 
-export enum TransferAirtimeExitNames {
```
rowanseymour, line 593: not used

src/components/flow/routers/airtime/__snapshots__/AirtimeRouterForm.test.ts.snap
```diff
@@ -156,7 +156,7 @@ Array [
             ],
             "category_uuid": "763e4844-3e1b-407a-a1b5-5fdfcd308b41",
             "type": "has_category",
-            "uuid": "1e47a1e1-3c67-4df5-adf1-da542c789adb",
+            "uuid": "e67ce051-f3e4-4c42-807b-e393df5fa382",
```
rowanseymour, line 159: cases uuids change which would break localization.. but these types of nodes shouldn't have localized case args

PR #1279, opened 2025-04-16
Reduce text max length to give room for expression evaluation

norkans7:


ericnewcomer:
So do we want to merge this or no?

rowanseymour:
I don't think we can.. would need to migrate existing flows

src/components/flow/actions/sendemail/SendEmailForm.tsx
```diff
@@ -14,8 +14,8 @@ import i18n from 'config/i18n';
 import { renderIssues } from '../helpers';
 
 const EMAIL_PATTERN = /\S+@\S+\.\S+/;
-const EMAIL_MAX_LENGTH_SUBJECT = 1000;
-const EMAIL_MAX_LENGTH_MESSAGE = 10000;
+const EMAIL_MAX_LENGTH_SUBJECT = 640;
```
rowanseymour, line 17: @norkans7 where is 640 coming from? engine has 1000

PR #1282, opened 2025-04-17
Enforce 10 QRs limit and MAX_TEXT_LEN for localizations message

norkans7:
requires https://github.com/nyaruka/temba-components/pull/490 for completeness


ericnewcomer:
Totally missed this, so working in your changes separately. Gah.

PR #1310, opened 2025-05-13
Fix creating new open ticket and airtime splits

rowanseymour:


norkans7 (approved):

src/components/flow/routers/ticket/helpers.ts
```diff
@@ -28,10 +28,14 @@ export const nodeToState = (settings: NodeEditorSettings): TicketRouterFormState
   const router = settings.originalNode.node.router as SwitchRouter;
 
   let note = { value: '' };
-  let resultName = { value: router.result_name || '' };
+  let resultName = { value: '' };
   let assignee: FormEntry = { value: null };
   let topic: FormEntry = { value: null };
 
+  if (router) {
+    resultName = { value: router.result_name };
```
norkans7, line 36: Now I think we need `resultName = { value: router.result_name || '' };` just in case the route has undefined result_name

PR #1324, opened 2025-06-24
Fix rendering of incoming messages in simulator

rowanseymour:


src/components/simulator/Simulator.tsx
```diff
@@ -368,6 +336,11 @@ export class Simulator extends React.Component<SimulatorProps, SimulatorState> {
         }
       }
 
+      // add msg we sent in as first event in sprint
+      if (msgEvt) {
+        runContext.events.unshift(msgEvt);
+      }
```
rowanseymour, line 342: need to include the `msg_received` event for the thing we just sent

PR #1329, opened 2025-07-23
Update simulation now that events have UUIDs, not message objects within events

rowanseymour:


src/components/simulator/Simulator.tsx
```diff
@@ -490,11 +490,11 @@ export class Simulator extends React.Component<SimulatorProps, SimulatorState> {
       const now = new Date().toISOString();
 
       const msgInEvt: any = {
+        uuid: createUUID(),
```
rowanseymour, line 493: should be v7 but doesn't matter since we don't persist simulated events

PR #1330, opened 2025-07-30
Fix simulator contact

norkans7:
Currently the simulator is only simulating with the flow main language even if the use expect the translation being views in the editor


rowanseymour:
Description please.. not sure what's broken here

PR #1335, opened 2025-08-20
Handle `run_ended` events instead of faking them

rowanseymour:


src/components/simulator/LogEvent.tsx
```diff
@@ -457,13 +457,18 @@ export default class LogEvent extends React.Component<EventProps, LogEventState>
         break;
       case 'webhook_called':
         return this.renderWebhook(Types.call_webhook);
-      case 'flow_entered': // deprecated
```
rowanseymour, line 460: @ericnewcomer this event was only generated for subflows whereas `run_started` now happens for all runs - so you see it at the start of every simulator session... which would be important in a scenario where the simulator is launched from some place other than the editor.

PR #1344, opened 2025-09-15
Don't look for localizable actions on router nodes

ericnewcomer:


src/components/translator/TranslatorTab.tsx
```diff
@@ -142,35 +142,37 @@ export class TranslatorTab extends React.Component<TranslatorTabProps, Translato
...
-                category,
-                this.props.localization
-              )
-            );
-          });
-        }
+          const typeConfig = getTypeConfig(getType(renderNode));
+
+          let translations: Translation[] = [];
+          if (this.state.translationFilters.categories) {
```
norkans7, line 150: This if block is repeated again and I think not needed

PR #1345, opened 2025-09-16
Remove unnecessary if block

norkans7:


src/components/translator/TranslatorTab.tsx
```diff
@@ -147,20 +147,19 @@ export class TranslatorTab extends React.Component<TranslatorTabProps, Translato
           const typeConfig = getTypeConfig(getType(renderNode));
 
           let translations: Translation[] = [];
-          if (this.state.translationFilters.categories) {
```
norkans7, line 150: was same if as on https://github.com/nyaruka/floweditor/pull/1345/files#diff-6b7dd5d315839b3886764a85e406f339b19090df6034ef565cc61503c29dc4abL146

